PR #25, opened 2017-02-12
TEST PR for seeing circle status

vanderhoop:


vanderhoop:
this was just a test of seeing the circle status, and we can indeed see it. woot! closing

PR #65, opened 2017-03-07
Fix compiler warnings

Zanadar:
Running migrations throws a few errors about naked variables.  This PR cleans those up.


vanderhoop:
MERGING

PR #72, opened 2017-03-12
Feature/submit action items toggle 43

tnewell5:
Hey Travis!

I completed the feature - whew!

I did run into an issue with one test that I was not able to solve. When I was trying to verify that the category dropdown was displayed on render for the Room component, I was getting a failing test. I tried it many ways, including the following, with no results:
```es6
expect(roomComponent.containsAllMatchingElements([
  <option value="happy">happy</option>,
  <option value="sad">sad</option>,
  <option value="confused">confused</option>
])).to.equal(true)
```

```es6
expect(roomComponent.containsMatchingElement(
  <IdeaSubmissionForm onIdeaSubmission={onSubmitIdeaStub} onToggleActionItem={onToggleActionItemStub} />
)).to.equal(true)
```

It did work just fine when I tested for the UserList component, which baffles me.
```es6
expect(roomComponent.containsMatchingElement(
  <UserList users={[]} />
)).to.equal(true)
```

Let me know if this is sufficient for the PR.

Thanks for all your help today!!
Tanya

### DIFF
--- a/web/static/js/components/idea_submission_form.jsx
+++ b/web/static/js/components/idea_submission_form.jsx
@@ -79,7 +79,6 @@ class IdeaSubmissionForm extends Component {
           </div>
           <div className="three wide field">
             <div className="ui toggle checkbox">
-              <input type="checkbox" />
               <input type="checkbox" onChange={ this.handleToggleChange } />
               <label>Toggle Action Items</label>
             </div>
--- a/web/static/js/components/idea_submission_form.jsx
+++ b/web/static/js/components/idea_submission_form.jsx
@@ -66,6 +66,13 @@ class IdeaSubmissionForm extends Component {
               </button>
             </div>
           </div>
+          <div className="three wide field">
+            <div className="ui toggle checkbox">
+              <input type="checkbox" />
+              <input type="checkbox" onChange={ this.props.onToggleActionItem } />
+              <label>Toggle Action Items</label>
+            </div>
+          </div>
         </div>
       </form>
     )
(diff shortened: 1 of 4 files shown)


web/static/js/components/idea_submission_form.jsx
```diff
@@ -26,26 +27,37 @@ class IdeaSubmissionForm extends Component {
...
   }
 
+  handleToggleShowCategories(event) {
+    this.setState({ showCategories: !this.state.showCategories })
+  }
+
+  handleToggleChange(event) {
+    this.props.onToggleActionItem()
+    this.handleToggleShowCategories()
+    this.setState({ category: "action-item"})
```
vanderhoop: hey @tnewell5. think there might be a small bug here. if we're setting the category to `"action-item"` regardless of the `showCategories` state, i think that if someone unchecks the action item toggle, all submitted ideas will go directly to the action items column until the user explicitly changes the select option to a new category. let me pull the branch and play around with it

vanderhoop: Pulled the branch and that's indeed the behavior, but I realize in retrospect that going back to the previously selected category wasn't in the acceptance criteria, so it doesn't block the PR. continuing in review

vanderhoop:
Nice work, Tanya. The only note I have, other than the one above, is that this introduces a few linter errors, mostly due to whitespace in the React components. I'll take care of them post-merge, but please run the linter via `mix lint` before any new PRs. Thanks!

PR #77, opened 2017-03-14
User Deletes ideas

Zanadar:
## Changes
- Users can delete ideas

** Note **
- Specs are passing, but kinda weak, IMO.  The blocking bug makes the testing fail for reasons unknown.  Running through the feature on my local works fine.


Zanadar:
The branch I'm merging into may need rebaing on master

Issue #54, opened 2017-02-26
enable email distribution of action items


vanderhoop:
As a retrospective facilitator who knows that action items are easily forgotten,
I would like to be able to email retro participants the action items that come out of a retro,
So that participants are reminded of the action items after leaving the retro.

__Pre-Reqs:__
  - #46 (only authorize the facilitator to shift the retro to the action items stage)

__Acceptance Criteria:__
  - [x] there's a SemanticUI "Labeled Icon Button" that reads "Send Action Items" to the right of the "action items" toggle from #43.
    - __Note:__ markup for button [supplied in comment below](https://github.com/vanderhoop/remote_retro/issues/54#issuecomment-282584364). 
  - [ ] within minutes of clicking the button, any retrospective participant who submitted an idea to the retro receives an email containing the action items from the retro.
  - [x] a javascript alert appears to the retrospective facilitator upon receiving a reply from the server
    - Success alert: "Consider it done. Participants will receive an email breakdown of the action items shortly."
    - Error alert: "It seems there was an error. Please try to send the action items again in a few minutes. Thanks for your patience."

__Helpful Docs:__
  - http://www.phoenixframework.org/docs/sending-email
  - http://semantic-ui.com/elements/button.html#labeled-icon-buttons
  - http://semantic-ui.com/elements/icon.html

vanderhoop:
__HTML:__
```html
<button class="ui right labeled teal icon button">
  Send Action Items
  <i class="send icon"></i>
</button>
```
__Produces:__
<img width="200" src="https://cloud.githubusercontent.com/assets/4386645/23343321/adc5c340-fc26-11e6-9291-b6e3ae9e71f9.png"/>

vanderhoop:
closing, as this was completed *months* ago, highlighting my preposterous oversight

PR #78, opened 2017-03-15
v1 working email to send action items #54

bglusman:
@vanderhoop should probably add tests to this, and the email formatting isn't really correct for plaintext vs html, but this works and sends, and locally shows email under a /sent_emails path

I have a sendgrid API key I can send you offline (also have a SparkPost API key, but it was giving me 404's for unclear reasons so just tried Sendgrid instead).


web/router.ex
```diff
@@ -1,6 +1,10 @@
 defmodule RemoteRetro.Router do
   use RemoteRetro.Web, :router
 
+  if Mix.env == :dev do
+    forward "/sent_emails", Bamboo.EmailPreviewPlug
```
vanderhoop: hmmm... what does this do exactly? i don't see this path referenced anywhere else? my guess is its boilerplate for Bamboo? Is the path prescriptive?

lib/remote_retro/emails.ex
```diff
@@ -0,0 +1,31 @@
...
+      html_body: html_retro_action_items(retro_id)
+    )
+  end
+
+  defp text_retro_action_items(retro_id) do
+    retro_action_items(retro_id)
+    |> Enum.join("\n")
+  end
+
+  defp html_retro_action_items(retro_id) do
```
vanderhoop: you da man!

vanderhoop:
Nice work, @bglusman! Can you add some tests around Bamboo and the new component? Happy to help if you want any! And if you don't have time, just let me know and I can add them myself this weekend. Thanks!

web/router.ex
bglusman: Yeah the route name is arbitrary but copied from example, this is basically like the rails letter_opener thing to preview sent mails locally rather than actually sending, similarly dev.exs has config to use LocalAdapter instead of real API, and this plug only works with local adapter.

lib/remote_retro/emails.ex
bglusman: Like I said, I don't actually think this is write/html for emails is different than web, and br's aren't really going to cut it, but presumably this basic approach will work with some tweaks.

bglusman:
@vanderhoop pretty busy getting ready for OpenPantry hack day this weekend (all invited!) so I might get to next week but if you get to this weekend, please go for it!

lib/remote_retro/emails.ex
```diff
@@ -0,0 +1,31 @@
+defmodule RemoteRetro.Emails do
+  import Bamboo.Email, except: [from: 2]
+  import Ecto.Query
+  alias RemoteRetro.Repo
+  def action_items_email(email_address, retro_id) do
+    new_email(
```
vanderhoop: todo: retrieve the action items from the database here, so that we don't hit the database once for the html email and once for the text email

web/channels/retro_channel.ex
```diff
@@ -7,6 +7,9 @@ defmodule RemoteRetro.RetroChannel do
   alias RemoteRetro.Presence
   alias RemoteRetro.PresenceUtils
   alias RemoteRetro.Idea
+  alias RemoteRetro.Emails
+  alias RemoteRetro.Mailer
+  alias RemoteRetro.Retro
```
vanderhoop: todo: no need for this alias

web/static/js/components/send_action_items_email.jsx
```diff
@@ -0,0 +1,38 @@
...
+  componentDidMount() {
+    this.props.retroChannel.on("email_send_status", (payload) => {
+      this.setState({ disabled: false })
+      alert(payload.success ? successMessage : errorMessage)
+    })
+  }
+
+  render() {
+    return (
+      <button className="ui right labeled teal icon button sendActionEmal" disabled={this.state.disabled} onClick={this.sendActionItemsEmail}>
```
vanderhoop: todo: delete unused `sendActionEmal` class

vanderhoop:
__Tasks Left__
- [ ] add unit test to RetroChannel for the handling of "send_action_items_email" event
  - `assert_broadcast` of map containing the success/failure value

bglusman:
Wow and I thought after we added some t are this was ready to merge 3 weeks ago!  Hope what I did helped anyway, sorry I haven't had time to do more since... Still hoping to recruit some Striders to help out on OpenPantry though!  Ran into Brett at a SpaceX panel last night, he suggested talking to Xander about it?  Hope you're all doing well!

PR #79, opened 2017-03-15
Feature/remove ideas

Zanadar:
## Changes
- Users can delete ideas

** Note **
- Specs are passing, but kinda weak, IMO.  The blocking bug makes the testing fail for reasons unknown.  Running through the feature on my local works fine.

### DIFF
--- a/test/features/retro_idea_realtime_update_test.exs
+++ b/test/features/retro_idea_realtime_update_test.exs
@@ -1,6 +1,24 @@
 defmodule RetroIdeaRealtimeUpdateTest do
   use RemoteRetro.IntegrationCase, async: false
 
+  test "deleting an idea from a list", %{session: session_one, retro: retro} do
+    {:ok, session_two} = Wallaby.start_session()
+
+    retro_path = "/retros/" <> retro.id
+    session_one = authenticate(session_one) |> visit(retro_path)
+    session_two = authenticate(session_two) |> visit(retro_path)
+
+    idea = %{category: "sad", body: "user stories lack clear business value"}
+    submit_idea(session_two, idea)
+
+    delete_idea(session_two, idea)
+    ideas_list_text = session_two |> find(Query.css(".sad.ideas", visible: false)) |> Element.text
+    refute String.contains?(ideas_list_text, idea.body)
+
+    Wallaby.end_session(session_one)
+    Wallaby.end_session(session_two)
+  end
+
   test "the immediate appearance of other users submitted' ideas", %{session: session_one, retro: retro} do
     {:ok, session_two} = Wallaby.start_session()
 
@@ -11,7 +29,7 @@ defmodule RetroIdeaRealtimeUpdateTest do
     ideas_list_text = session_one |> find(Query.css(".sad.ideas", visible: false)) |> Element.text()
     refute String.contains?(ideas_list_text, "user stories lack clear business value")
 
-    submit_idea(session_two, %{ category: "sad", body: "user stories lack clear business value" })
+    submit_idea(session_two, %{category: "sad", body: "user stories lack clear business value" })
 
     ideas_list_text = session_one |> find(Query.css(".sad.ideas")) |> Element.text
     assert String.contains?(ideas_list_text, "user stories lack clear business value")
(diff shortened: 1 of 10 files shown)


web/static/js/components/room.jsx
```diff
@@ -24,12 +26,33 @@ class Room extends Component {
...
       this.setState({ ideas: [...this.state.ideas, newIdea] })
     })
+
+    this.props.retroChannel.on("idea_deleted", idea => {
+      this.setState({ ideas: this._removeIdea(this.state.ideas, idea.id)})
+    })
+  }
+
+  _removeIdea(ideas, id) {
+    let index = ideas.map(idea => idea.id).indexOf(id)
```
ianmcnally: a simpler approach might be to use a filter,

```javascript
return ideas.filter(idea => idea.id !== id)
```

would remove the `-1` check, which is hard to read, and the slice-a-roni.

web/static/js/components/room.jsx
```diff
@@ -13,7 +13,9 @@ class Room extends Component {
     super(props)
     this.state = { ideas: [], showActionItem: false }
     this.handleIdeaSubmission = this.handleIdeaSubmission.bind(this)
+    this.handleIdeaDeletion = this.handleIdeaDeletion.bind(this)
     this.handleToggleActionItem = this.handleToggleActionItem.bind(this)
+    this._removeIdea = this._removeIdea.bind(this)
```
ianmcnally: might want to migrate to [class property initializers](https://babeljs.io/docs/plugins/transform-class-properties/), .e.g.,

```javascript
class ThisComponent {
  _removeIdea = () => {
    // sweet code here
  };
}

makes binding less a chore. can also move `state` initialization to an initializer and completely scrap the constructor. [proof](https://twitter.com/ianmcnally_/status/780523739242962944)

File: web/static/js/components/room.jsx [hunk 3]
```diff
@@ -38,9 +61,9 @@ class Room extends Component {
     return (
       <section className={styles.wrapper}>
         <div className={`ui equal width padded grid ${styles["category-columns-wrapper"]}`}>
```
ianmcnally: unrelated to your change, but:
if you pass css loader an [option](https://github.com/webpack-contrib/css-loader#camelcase) to camel case class names, you could write styles.categoryColumnsWrapper.

web/static/js/components/room.jsx
```diff
@@ -13,7 +13,9 @@ class Room extends Component {
     super(props)
```
ianmcnally: oh other random note (sry for spam). the community's moving away from the `jsx` file extension, in favor of good ol `.js`.

web/static/js/components/room.jsx
ianmcnally: did this in #80

vanderhoop:
hey @Zanadar, debugging the Wallaby weirdness as we speak. the only other blocking feedback i have is the snake_case parameter that @ianmcnally mentioned, and that it'd be great if we had a test around the function passed as a prop to the `IdeaListItem` being invoked on click of the button.

test/features/retro_idea_realtime_update_test.exs
```diff
@@ -1,6 +1,24 @@
...
+  test "deleting an idea from a list", %{session: session_one, retro: retro} do
+    {:ok, session_two} = Wallaby.start_session()
+
+    retro_path = "/retros/" <> retro.id
+    session_one = authenticate(session_one) |> visit(retro_path)
+    session_two = authenticate(session_two) |> visit(retro_path)
+
+    idea = %{category: "sad", body: "user stories lack clear business value"}
+    submit_idea(session_two, idea)
+
```
vanderhoop: hey @Zanadar. poked around with this test for over an hour, and got false positives when disabling the push from the front end. Did some debugging, and found that the callback passed as a prop _does_ get invoked, but the function that that callback invokes isn't fired. 

Here's the output in Chrome:

screen shot 2017-03-18 at 5 21 06 pm


And here's the output from the e2e test that deletes an idea:

image
screen shot 2017-03-18 at 5 21 06 pm

I think the issue is that the `Room` component's `handleIdeaDeletion` method uses `.bind(this)`, and historically, bind and phantomjs haven't played well together. Gonna play around a bit more, but wanted to post this to see if it gave you any ideas.

vanderhoop:
hey @Zanadar, rebased master onto `feature/deleting-submitted-ideas` and cherry-picked your commits onto it afterwards. gonna make this a facilitator-only feature and get it into master. closing!

PR #80, opened 2017-03-15
Use camel case formatting for css module imports

ianmcnally:
In #79, it came upon mine very eyes an opportunity to make importing a styles object a little easier by adding the camelCase option.

E.g.,
__before__
```javascript
import styles from './thing.css'

styles['this-thing']
```

__after__
```javascript
import styles from './thing.css'

styles.thisThing
```

Also took it as an opportunity to use the alternative loader syntax, which I hope improves readability.


ianmcnally:
🙏we got green. ran into some strange node module path problems that _i think_ were unrelated.

vanderhoop:
hey @ianmcnally this is great! I had no idea their was a `camelCase` config option, and I'm glad there is! The CircleCI path stuff is kinda wonks, and I'm not sure why it would break for your branch and not others. We ran into some stuff early on in the way `yarn add global` installs executables compared to `npm install -g`, and Circle didn't react kindly to it, so we reverted to npm for global installs.

I'm gonna cherry-pick the functional commits and do a time-boxed experiment to see if I can't get it to work without the circle changes. If that fails, merging!

ianmcnally:
yas

PR #81, opened 2017-03-17
Move app-specific prop types to own file

ianmcnally:
Helps with:
- Ensuring the contract, via props, for domain-specific objects and data types
- DRY as the desert
- This thing called testing

This pattern helped a lot at a couple Stride clients 👌.


vanderhoop:
hey @ianmcnally, love this. so simple. Merging, but I'm going to rename the `prop-types` file to `prop_types`, in keeping with the rest of the application convention.

PR #83, opened 2017-03-20
Remove ActionItemToggle for non-facilitators

crq:


test/components/room_test.js
```diff
@@ -59,12 +75,12 @@ describe("Room component", () => {
...
 
       expect(roomComponent.state("ideas")).to.eql([
-        { arbitrary: "content" }
+        { arbitrary: "content" },
       ])
     })
 
     it("pushes the value passed with `new_idea_received` into the `ideas` array", () => {
-      roomComponent.setState({ ideas: [{ body: "first idear" }]})
+      roomComponent.setState({ ideas: [{ body: "first idear" }] })
```
vanderhoop: thanks for the cleanup here, @crq

vanderhoop:
nice work, @crq. merging!

test/components/remote_retro_test.js
```diff
@@ -0,0 +1,48 @@
...
+      userToken: {
+        user: { facilitator: true },
+      },
+    }
+
+    it("renders a room with isFacilitiator true", () => {
+      const wrapper = shallow(
+        <RemoteRetro userToken="userToken" retroChannel={mockRetroChannel} />)
+
+      wrapper.setState({ presences })
```
ianmcnally: setState seems brittle. could this setup follow the user flow?

crq: Travis and I have been talking about moving away from setting state in tests. We'd probably want to introduce something like redux to handle the channel callbacks. Currently different callbacks are being declared to set state on several difference components. Ideally we'd change applications state and pass the props only to components that need to react to the change.

ianmcnally: respect.

looking at how presence gets set, you could mock `props.retroChannel.on`, and call the second arg with presences. that said, thinking about a longer term approach sounds like a better idea. you the man, @crq. don't let anyone tell you otherwise.

PR #90, opened 2017-03-22
Only Facilitator can enable/disable action items

crq:


vanderhoop:
this is great, @crq. gonna update some of the test language server-side, but overall really nice work!

Issue #52, opened 2017-02-23
attribute an idea to a user


vanderhoop:
As a retrospective participant,
I would like to know who submitted an idea,
So that I can ask them to elaborate on their idea.

__Acceptance Criteria:__

- [ ] ideas on the board are prefixed with `"<FIRST NAME>: "` where first name is the first name of whoever submitted the idea :)
- [ ] when a user enters a retro that already has ideas on the board, they can see the name of whoever submitted a particular idea
  - __Note:__ this involves persistence

__Example:__

```
┌─────────────┐
│      :)     │
├─────────────┤
│ sean: linter│
│ dan: linter!│
│ meg: linter!|
└─────────────┘
```

caporta:
@vanderhoop I don't believe Acceptance Criterion #2 involves persistence now that we've gone the `presences[userToken]` route, but I could be wrong. Tested with ngrok and everything looks gravy.

PR #93, opened 2017-03-29
Feature/attribute idea to user 52

caporta:
@vanderhoop Season's greetings!

Here's a PR for #52, which focuses on tethering an author to an idea. @peterylai was also a core contributor to this feature.

One design concession we made was passing the entire `currentUser` object down the component chain as a prop; we would have preferred to evaluate `currentUser.user.given_name` much further upstream, but due to the rendering issues we've discussed (e.g. multiple `render` invocations on the same component, with prop values as `undefined` on first render), this would have required conditional logic that felt a bit dirty. This will hopefully be be solved upon our move to Redux.

### DIFF
--- a/web/channels/retro_channel.ex
+++ b/web/channels/retro_channel.ex
@@ -9,7 +9,7 @@ defmodule RemoteRetro.RetroChannel do
   alias RemoteRetro.Idea
 
   def join("retro:" <> retro_id, _, socket) do
-    query = from idea in Idea, select: map(idea, [:body, :category, :inserted_at, :retro_id, :id]), where: idea.retro_id == ^retro_id
+    query = from idea in Idea, where: idea.retro_id == ^retro_id
     existing_ideas = Repo.all(query)
     socket_ideas = Phoenix.Socket.assign(socket, :ideas, existing_ideas)
     socket_retro = Phoenix.Socket.assign(socket_ideas, :retro_id, retro_id)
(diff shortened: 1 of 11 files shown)


web/channels/retro_channel.ex
```diff
@@ -9,7 +9,7 @@ defmodule RemoteRetro.RetroChannel do
   alias RemoteRetro.Idea
 
   def join("retro:" <> retro_id, _, socket) do
-    query = from idea in Idea, select: map(idea, [:body, :category, :inserted_at, :retro_id, :id]), where: idea.retro_id == ^retro_id
+    query = from idea in Idea, where: idea.retro_id == ^retro_id
```
vanderhoop: love this

web/static/js/components/remote_retro.jsx
```diff
@@ -34,6 +34,7 @@ class RemoteRetro extends Component {
 
     return (
       <Room
+        currentUser={currentPresence}
```
vanderhoop: the misalignment in naming here makes me think we'll want to go with `currentPresence` as the prop name. that way, , up above, we won't introducee the weirdness of calling `currentUser.user`

vanderhoop:
hey @caporta! overall, this looks great. two quick things before getting this in:
  1. please rename the `currentUser` prop to `currentPresence`
  1. remove the WIP commit from history!

web/static/js/components/remote_retro.jsx
caporta: ✔️

caporta:
@vanderhoop yo! This should be good to go

vanderhoop:
nice work, @caporta and @peterylai! mergin'

PR #101, opened 2017-04-14
Refactor/build store with user reducer and action 99

caporta:
@vanderhoop 👀

### DIFF
--- a/test/actions/user_actions_test.js
+++ b/test/actions/user_actions_test.js
@@ -1,12 +1,11 @@
-import { expect } from 'chai'
+import { expect } from "chai"
 
-import * as actionCreators from '../../web/static/js/actions/user'
+import * as actionCreators from "../../web/static/js/actions/user"
 
-describe('addUser', () => {
-  it('should create an action to add user to users list', () => {
-    const user = { given_name: 'Tiny Rick' }
-    const expectedAction = { type: 'ADD_USER', user }
+describe("addUser", () => {
+  const user = { given_name: "Tiny Rick" }
 
-    expect(actionCreators.addUser(user)).to.deep.equal(expectedAction)
+  it("should create an action to add user to users list", () => {
+    expect(actionCreators.addUser(user)).to.deep.equal({ type: "ADD_USER", user })
   })
 })
--- a/test/actions/user_actions_test.js
+++ b/test/actions/user_actions_test.js
@@ -1,0 +1,12 @@
+import { expect } from 'chai'
+
+import * as actionCreators from '../../web/static/js/actions/user'
+
+describe('addUser', () => {
+  it('should create an action to add user to users list', () => {
+    const user = { given_name: 'Tiny Rick' }
+    const expectedAction = { type: 'ADD_USER', user }
+
+    expect(actionCreators.addUser(user)).to.deep.equal(expectedAction)
+  })
+})
--- a/test/reducers/user_reducer_test.js
+++ b/test/reducers/user_reducer_test.js
@@ -1,41 +1,37 @@
-import { expect } from 'chai'
+import { expect } from "chai"
 
-import userReducer from '../../web/static/js/reducers/user'
+import userReducer from "../../web/static/js/reducers/user"
 
+describe("user reducer", () => {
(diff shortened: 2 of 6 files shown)


test/reducers/user_reducer_test.js
```diff
@@ -0,0 +1,41 @@
+import { expect } from 'chai'
+
+import userReducer from '../../web/static/js/reducers/user'
+
+
+describe('user reducer', () => {
+  describe('when no state is passed', () => {
+    it('should return the initial state', () => {
+      expect(userReducer(undefined, {})).to.deep.equal([])
```
vanderhoop: kinda hard to tell what the initial state is in this test. could you make it explicit via the `describe` or `it` descriptions?

test/reducers/user_reducer_test.js
```diff
@@ -0,0 +1,41 @@
...
+
+      expect(userReducer([], action)).to.deep.equal(expectedState)
+    })
+  })
+
+  describe('when there are existing users', () => {
+    it('should handle ADD_USER', () => {
+      const user = { given_name: 'Tiny Rick' }
+      const action = { type: 'ADD_USER', user }
+      const expectedState = [{ given_name: 'Morty' }, user]
```
vanderhoop: would love to avoid the expectation-as-variable in our unit tests. just put it in there inline. 

caviat: if you're going to use the variable in multiple places, it's fine to keep it a variable

web/static/js/app.js
```diff
@@ -2,9 +2,12 @@
 
 import React from "react"
 import { render } from "react-dom"
+import { createStore } from "redux"
+import { Provider } from "react-redux"
 
 import RemoteRetro from "./components/remote_retro"
 import RetroChannel from "./services/retro_channel"
+import rootReducer from "./reducers/index"
```
vanderhoop: i think (think) that you can leave off the `/index` here, as `index.js` is the default file that will be imported if you `import` a directory

web/static/js/reducers/index.js
```diff
@@ -0,0 +1,10 @@
+import { combineReducers } from 'redux'
+
+import user from './user'
+
+
```
vanderhoop: lot of extra whitespace in these files, especially after imports. only one line is necessary

test/actions/user_actions_test.js
```diff
@@ -0,0 +1,12 @@
+import { expect } from 'chai'
+
+import * as actionCreators from '../../web/static/js/actions/user'
+
+describe('addUser', () => {
+  it('should create an action to add user to users list', () => {
+    const user = { given_name: 'Tiny Rick' }
+    const expectedAction = { type: 'ADD_USER', user }
+
+    expect(actionCreators.addUser(user)).to.deep.equal(expectedAction)
```
vanderhoop: same comment as below. try to stay away from pulling expectations out into variables, as it's easier to read if they're inline

vanderhoop:
hey @caporta and @peterylai, nice work! made some nitpicky comments. but aside from those, make sure you run the linter, because it's going to puke big time with all those single-quote strings. after that, we're mergeable.

test/reducers/user_reducer_test.js
caporta: ✅

test/reducers/user_reducer_test.js
caporta: :check

web/static/js/app.js
caporta: ✅

web/static/js/reducers/index.js
caporta: ✅

test/actions/user_actions_test.js
caporta: ✅

caporta:
@vanderhoop PR has been cleaned up & de-linted as per the requests above, with one small exception. The linter is flagging `Prefer default export` in `actions/user.js`, but this should only be temporary, as this rule won't apply when new action creators `e.g. { type: 'REMOVE_USER' }` are added in the very near future. The current implementation reflects idiomatic Redux.

vanderhoop:
you da man!

Issue #102, opened 2017-04-19
persist users who authenticate from google


vanderhoop:
As a developer,
I would like to persist proper user records in the database,
So that I properly associate a `user`s with the `idea`s they've created and the `retro`s they've joined.

__By Way of Explanation:__

To this point, we've punted on adding a full-fledged `user` model to the domain, instead leveraging in-memory attributes from the session (all of which are just Google OpenId attributes) when needed. For example, the `ideas` table has an `author` attribute that's a string, rather than a proper foreign key. This has been adequate for now, but in upcoming features, we're going to need to associate records with users in a more robust way.

__Acceptance Criteria:__

- [x] there's a `users` table that can persist the following datastructure as a single row.
  - __Note:__ the map below represents a subset of all the attributes passed to us by Google OpenID. I kept the field names in parity with the OpenId field names.

```elixir
%{
  "email" => "mistertestuser@gmail.com",
  "family_name" => "Vander Hoop",
  "gender" => "male",
  "given_name" => "Travis",
  "locale" => "en",
  "name" => "Test User",
  "picture" => "https://lh6.googleusercontent.com/-cZI40d8YpIQ/AAAAAAAAAAI/AAAAAAAAABs/gmDI7LQ2Lo0/photo.jpg?sz=50",
  "profile" => "https://plus.google.com/108658712426577966861",
}
```
- [x] when a user [authenticates via Google OAuth](https://github.com/stride-nyc/remote_retro/blob/22c3d4ff92167cee1fe861cbadd21ae9ba36c560/web/controllers/auth_controller.ex#L9-L13) for the first time in our application, a user record is created in the `users` table, and this user is placed on the session.
- [x] when a user authenticates via Google OAuth, and that user is already represented in our database, a new record is *not* created, and the existing user is placed on the session.
  - the `email` attribute coming from Google is unique, so that should be a good check.
- [x] there are unit tests

__Helpful Links:__

- [AuthController Action That Catches Google OpenId Info](https://github.com/stride-nyc/remote_retro/blob/22c3d4ff92167cee1fe861cbadd21ae9ba36c560/web/controllers/auth_controller.ex#L9-L13) 
- https://hexdocs.pm/ecto/Ecto.Repo.html
- https://hexdocs.pm/ecto/Ecto.Migration.html

vanderhoop:
implemented

Issue #74, opened 2017-03-13
E2E specs + wifi issues


Zanadar:
When Wifi is off, I observe two things:
- Running e2e specs is significantly faster (18 second vs 4 second reported in logs).
- Test failures are different.  In the case of WiFi being on, I get
`** (Wallaby.QueryError) Expected to find 1, visible element that matched the css '.sad.ideas' but 0, visible elements were found.`
With WIFI off I get:
`  Expected false or nil, got true`

This may be a wallaby bug, but I think the timing has something to do with auth going out to the internet where it doesn't need to?

vanderhoop:
hey @Zanadar I think this is owed to [the call for Semantic UI, which is hosted on a CDN](https://github.com/vanderhoop/remote_retro/blob/59db05708c1df9ab8efdd93b67084fc815f4ea26/web/templates/layout/app.html.eex#L11) (cloudflare). I had a commit that, but had to revert it due to it calling on relative paths for some icons

Zanadar:
Gonna close this as a wontfix

vanderhoop:
removed won't fix and reopening this, as it's a real issue for devs without a network connection

vanderhoop:
handled by [this commit](https://github.com/stride-nyc/remote_retro/commit/165001986f217c8c7cd1e3e3280d4541fd458d9f)

PR #104, opened 2017-04-21
User persistence

pulsedemon:
Addresses #102 
## Changes:
- Creates a Users table
- OAuthing initially creates the user (if they're new) and puts them into the session
- Check in new `hex.lock` (compatible with new hex v 0.16.0)


test/oauth/google_test.exs
```diff
@@ -1,6 +1,6 @@
 defmodule RemoteRetro.GoogleTest do
   use ExUnit.Case, async: true
-  alias RemoteRetro.Google
+  alias RemoteRetro.OAuth.Google
```
vanderhoop: thanks! this is far more idiomatic

web/router.ex
```diff
@@ -11,9 +11,7 @@ defmodule RemoteRetro.Router do
     plug :fetch_flash
     plug :protect_from_forgery
     plug :put_secure_browser_headers
-    if Application.get_env(:remote_retro, :add_mock_user_to_session) do
-      plug :add_mock_user_to_session
-    end
+
```
vanderhoop: @pulsedemon @Zanadar awesome change (I think?), but not quite sure how it works. don't you get redirected to authenticate with google if the user isn't on session? Or am I missing something?

web/router.ex
```diff
@@ -11,9 +11,7 @@ defmodule RemoteRetro.Router do
     plug :fetch_flash
     plug :protect_from_forgery
     plug :put_secure_browser_headers
-    if Application.get_env(:remote_retro, :add_mock_user_to_session) do
-      plug :add_mock_user_to_session
```
vanderhoop: because the `add_mock_user_to_session` is no longer used anywhere, let's delete the implementation as well (further down the file)

web/router.ex
pulsedemon: @vanderhoop Yeah, we thought it would be better to always authenticate if there isn't an active session.  Is that an issue?

priv/repo/migrations/20170419210307_create_user.exs
```diff
@@ -0,0 +1,15 @@
+defmodule RemoteRetro.Repo.Migrations.CreateUser do
+  use Ecto.Migration
+
+  def change do
+    create table(:users) do
+      add :google_user_info, :jsonb
```
vanderhoop: any reason for storing the user attributes as json? wouldn't it be easier to just cast the map to a row in the database, i.e. with a `given_name` column, `family_name` column, and so on?

config/dev.exs
```diff
@@ -44,7 +44,7 @@ config :remote_retro, RemoteRetro.Mailer, adapter: Bamboo.LocalAdapter
 
 config :remote_retro, :oauth_client, OAuth2.Client
 
-config :remote_retro, :add_mock_user_to_session, true
+config :remote_retro, :add_mock_user_to_session, false
```
vanderhoop: I believe this is the only place this config is referenced in the app, so we can delete this line outright

web/controllers/auth_controller.ex
```diff
@@ -1,13 +1,26 @@
...
+  alias RemoteRetro.User
 
   def index(conn, _params) do
     redirect conn, external: authorize_url!()
   end
 
   def callback(conn, %{"code" => code}) do
-    conn = put_session(conn, :current_user, Google.get_user_info!(code))
+    user_info = Google.get_user_info!(code)
+    conn = put_session(conn, :current_user, user_info)
```
vanderhoop: this one is a bit weird for me given that we're saying the session's `current_user` is really just the map of `google_user_info`, when in actuality our user model is a map that happens to have an attribute called `google_user_info` that has the same fields. Makes me feel more strongly that the `users` table should just have a column for each attribute.

web/models/user.ex
```diff
@@ -0,0 +1,21 @@
+defmodule RemoteRetro.User do
+  use RemoteRetro.Web, :model
+
+  schema "users" do
+    field :google_user_info, :map
+    field :email, :string
+    field :last_login, Ecto.DateTime
```
vanderhoop: didn't think of this, but great inclusion.

web/router.ex
vanderhoop: i believe so, yeah. to actually authenticate, Google needs to be able to POST to a specific endpoint with the appropriate `code` param, and they can't make POST requests to your `localhost` unless you expose a port via ngrok. So this `add_mock_user_to_session` plug is only used in development.

vanderhoop: and i don't want a network connection + ngrok to be necessary for devs to work on the project

pulsedemon: @vanderhoop We got it working on localhost without doing anything like that.  If you test it out with the credentials I pasted in slack you can see for yourself.

priv/repo/migrations/20170419210307_create_user.exs
pulsedemon: @vanderhoop I personally feel like jsonb makes sense because the data returned from Google shouldn't be editable and it also might change in the future, especially if we add more scopes.  Storing the data unstructured makes it simple to store any extra data in the future without modifying the schema.  The only reason we made email it's own field is because we needed a unique index.

web/router.ex
vanderhoop: pulled the branch, and you're right. Not sure how google is posting to localhost, but I won't question it.

One drawback here is that this adds a network dependency for devs who don't have a user on session. In truth, there's *already* a network dependency due to using a CDN-hosted semantic-ui stylesheet, but that's something I was hoping to kill, and I think @Zanadar  [was too](https://github.com/stride-nyc/remote_retro/issues/74). Let's connect on monday and discuss the pluses and minuses of actually making devs authenticate.

priv/repo/migrations/20170419210307_create_user.exs
vanderhoop: make sense. the ability to catch any additional fields that google sends in the future is certainly a benefit.

the drawback to this approach alone is that it bypasses the data integrity we could ensure using actual columns. For example, let's say in late 2017 google  changes the "given_name" field to "first_name" (unlikely, but possible). All users who joined prior to that change would see the proper value displayed when `user.google_user_info.given_name` is referenced downstream, but users who join after that change wouldn't see a name, and as developers we'd never know. Using columns with NOT NULL constraints (in conjunction with the unstructured JSONB storage) would alert us when we try to store a value that the google user info no longer provides.

A hybrid approach of makes the most sense to me, as it would provide us with:

  - comprehensive storage of all google data in the unstructured JSONB column
  - data integrity, ensuring that:
    - the fields we reference in the app exist
    - the server would have a catch-point for google changing fields we rely on
  - downstream, we wouldn't have have to reference core attributes at different levels of nesting
    - it would be kinda weird to say the id is 
    - i.e. we can say `user.given_name`, rather than `user.google_user_info.given_name`

The only drawback is some data duplication, but that's a small price to pay.

pulsedemon: I didn't think about that, but that makes sense and I agree with you.  Will make some changes

web/controllers/auth_controller.ex
```diff
@@ -9,10 +9,19 @@ defmodule RemoteRetro.AuthController do
 
   def callback(conn, %{"code" => code}) do
     user_info = Google.get_user_info!(code)
-    conn = put_session(conn, :current_user, user_info)
-
     user = Repo.get_by(User, email: user_info["email"])
-    user_params = %{email: user_info["email"], google_user_info: user_info, last_login: DateTime.utc_now}
+
+    user_params = %{
```
vanderhoop: rather than assign all of the attributes of `user_info` into a new data structure one by one, let's use `Map.merge` to produce a new map containing all the attributes of `user_info` along with the new attributes (`google_user_info`, `email`, and `last_login`)

vanderhoop:
hey @pulsedemon and @Zanadar. Three blocking changes on this guy, two of them minor, one of them a bit more thinky.

1. Use `Map.merge` rather than in-lining the copying of attributes (see [comment above](https://github.com/stride-nyc/remote_retro/pull/104#discussion_r112824179))
1. Add not-null constraints to the individual columns on the `users` table
   - I know that there are `validate_required` calls in our changeset, but I want to ensure we have protections at the database level as well.
1. Solve the problem of the environmental variables necessary to use Google OAuth in dev:
   - if you start a brand new shell without the environmental variables you sent me in slack, OAuth dev b0rks. This is problematic in that we don't want to have any old contributor having access to our `GOOGLE_OAUTH_CLIENT_SECRET` env var. 

__Where GOOGLE_OAUTH_CLIENT_SECRET is nil:__
image

__This happens:__

image

___

My personal feeling is that we should re-instate the `add_mock_user_to_session` plug in dev, as this avoids the problem of the env-var dependency and additionally keeps us from adding a network-connection dependency in dev. That said, you two are right that the `add_mock_user_to_session` plug does make working with the auth controller a bit odd at first (because you never really access it), so if you two can find an elegant solution, I'm all ears!

Zanadar:
Heya!  So we were chatting and think that making devs create OAUTH tokens and export them locally would be totally fine (we don't all have to share tokens). We're going to add a mix task to give devs a notice if they don't have them how to create and add something to the readme.

RE the network dependency... I kind a feel like this is a web app, so it probably OK to have it depend on being on connected to the internet. In any case, once somebody is authed their token will be good on session for a while so it should be fine if they want to work offline some of the time.

RE the names of the ENV_VARS, we were thinking that eg. `REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_ID` would be a good way to export them, avoiding clashes with other Oauth tokens that devs might have?

Thoughts @vanderhoop ?

vanderhoop:
yo @Zanadar and @pulsedemon. when will the mix task be invoked? is this check run every time we try to run the default `mix` task (as a subtask?)

vanderhoop:
+1 on the naming of `REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_ID`

the network dependency issue is likely to bite some folks, as I believe they'll have to re-authenticate  anytime they quit their browser, but let's roll with it for now and see if anyone complains.

vanderhoop:
@Zanadar @pulsedemon Just pulled the branch, and it's beautiful. Nice work, you two! Mergin!

PR #106, opened 2017-04-21
added user icons to user_list_item

pulsedemon:
^^

### DIFF
--- a/test/components/user_list_item_test.js
+++ b/test/components/user_list_item_test.js
@@ -16,6 +16,11 @@ describe("UserListItem", () => {
       const wrapper = shallow(<UserListItem user={user} />)
       expect(wrapper.text()).to.match(/dylan$/i)
     })
+
+    it("renders a list item that does not have a picture", () => {
+      const wrapper = shallow(<UserListItem user={user} />)
+      expect(wrapper.find('i.icon').html())
+    })
   })
 
   describe("passed a facilitator user", () => {
--- a/test/components/user_list_item_test.js
+++ b/test/components/user_list_item_test.js
@@ -40,15 +40,15 @@ describe("UserListItem", () => {
       picture: "http://some/image.jpg"
     }
 
-    it("renders a list item that does have a picture", () => {
+    it("renders a list item that has a picture", () => {
       const wrapper = shallow(<UserListItem user={user} />)
-      expect(wrapper.find('img.picture').html())
+      expect(wrapper.find('img.picture')).to.have.length(1)
     })
 
-    it("renders a list item that does not have a picture", () => {
+    it("renders a list item that displays an icon when a picture isn't present", () => {
       user.picture = '';
       const wrapper = shallow(<UserListItem user={user} />)
-      expect(wrapper.find('i.icon').html())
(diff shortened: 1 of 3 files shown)


web/static/js/components/user_list_item.jsx
```diff
@@ -8,13 +8,22 @@ function UserListItem(props) {
...
-        <div className="ui row">{ userName }</div>
-        <i className="huge user icon" />
+         {IconTag(props.user)}
+         <div className="ui row">{ userName }</div>
       </div>
     </li>
   )
 }
 
+const IconTag = function(user) {
```
vanderhoop: this is a great change. thanks for taking it on unprompted. only blocker for merging is the need for a couple unit tests to the `UserListItem` component?

pulsedemon: @vanderhoop Added 2 tests

web/static/js/components/user_list_item.jsx
```diff
@@ -8,13 +8,22 @@ function UserListItem(props) {
...
+  if (user.picture) {
+    return <img className={`${styles.picture}`} src={user.picture} />
```
vanderhoop: no need to interpolate here for providing the `className` prop. you could just reference `styles.picture` outright

web/static/js/components/user_list_item.jsx
```diff
@@ -8,13 +8,22 @@ function UserListItem(props) {
...
+  }
+  else {
```
vanderhoop: please lift the `else` onto the same line as the closing of the `if` block. pretty sure this will make the linter puke.

test/components/user_list_item_test.js
```diff
@@ -30,4 +31,24 @@ describe("UserListItem", () => {
...
+      is_facilitator: true,
+      picture: "http://some/image.jpg"
+    }
+
+    it("renders a list item that does have a picture", () => {
+      const wrapper = shallow(<UserListItem user={user} />)
+      expect(wrapper.find('img.picture').html())
+    })
+
+    it("renders a list item that does not have a picture", () => {
```
vanderhoop: let's update the test language, as what this is really doing is displaying an icon when the user lacks's a picture.

test/components/user_list_item_test.js
```diff
@@ -30,4 +31,24 @@ describe("UserListItem", () => {
...
+    const user = {
+      given_name: "treezy",
+      online_at: 803,
+      is_facilitator: true,
+      picture: "http://some/image.jpg"
+    }
+
+    it("renders a list item that does have a picture", () => {
+      const wrapper = shallow(<UserListItem user={user} />)
+      expect(wrapper.find('img.picture').html())
```
vanderhoop: also, these tests will always pass, because there are no matcher being used with the `expect` invocation. this could easily be like `expect(wrapper.find('img.picture')).to.equal(true)`

vanderhoop: correction: these would only pass *if* enzyme is able to `find` an image tag, but i'd rather be explicit in the expectation/matcher combo than just relying on expect

vanderhoop:
mergin!

PR #114, opened 2017-04-25
refactored preflight.ex and added new function to enforce installatio…

pulsedemon:
…n of specific hex version

@Zanadar @vanderhoop Let me know what you guys think about this

### DIFF
--- a/lib/tasks/preflight.ex
+++ b/lib/tasks/preflight.ex
@@ -8,7 +8,7 @@ defmodule Mix.Tasks.Preflight do
     check_hex_version()
   end
 
-  def check_envars do
+  defp check_envars do
     required_envars = [
       "REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_ID",
       "REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_SECRET",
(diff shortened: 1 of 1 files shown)


lib/tasks/preflight.ex
```diff
@@ -18,4 +23,17 @@ defmodule Mix.Tasks.Preflight do
       System.halt(:abort)
     end
   end
+
+  def check_hex_version do
+    required_version_num = "0.16.0"
+
+    hex_info = "mix hex.info | grep Hex" |> String.to_char_list |> :os.cmd |> List.to_string
```
vanderhoop: We can leverage `System.cmd/2` here to avoid the char-list transforms, and `String.contains?/2` to avoid the regex stuff

```elixir
{ hex_version_output, _ } = System.cmd("mix", ["hex", "--version"])
String.contains?(hex_version_output, required_version_num)
```

lib/tasks/preflight.ex
```diff
@@ -18,4 +23,17 @@ defmodule Mix.Tasks.Preflight do
       System.halt(:abort)
     end
   end
+
+  def check_hex_version do
```
vanderhoop: let's make these private via `defp`

vanderhoop:
i like the idea behind this; nice work. 3 small things:

- `def` -> `defp` for methods that aren't leveraged publicly
- simplify the version comparison logic (see comment above)
- add a "Please" to the head of the string we pass to `IO.puts` :)

lib/tasks/preflight.ex
pulsedemon: @vanderhoop Thanks!  Still new to elixir, so I'm not always sure what the best way to do things is.

vanderhoop:
mergin, with the caveat that we'll likely have to update this when they upgrade to hex 0.17.0

PR #117, opened 2017-04-25
add dynamic placeholder text

pulsedemon:


vanderhoop:
nice work, @qlaire/@pulsedemon. i think, rather than managing the changes of an additional attribute on state, we can just use the state's `category` value as the key into the `PLACEHOLDER_TEXTS` constant, like `placeholder={PLACEHOLDER_TEXTS[this.state.category]}`

web/static/js/components/idea_submission_form.jsx
```diff
@@ -16,7 +28,9 @@ class IdeaSubmissionForm extends Component {
   componentWillReceiveProps(nextProps) {
     if (nextProps.showActionItem !== this.props.showActionItem) {
       const category = nextProps.showActionItem ? "action-item" : this.defaultCategory
-      this.setState({ category })
+      this.setState({
```
vanderhoop: no need to multi-line this

web/static/js/components/idea_submission_form.jsx
```diff
@@ -37,7 +51,9 @@ class IdeaSubmissionForm extends Component {
   }
 
   handleCategoryChange(event) {
-    this.setState({ category: event.target.value })
+    this.setState({
```
vanderhoop: no need to multi-line this, now that we are only setting the one property

web/static/js/components/idea_submission_form.jsx
```diff
@@ -73,7 +89,7 @@ class IdeaSubmissionForm extends Component {
                 ref={input => { this.ideaInput = input }}
                 value={this.state.body}
                 onChange={this.handleIdeaChange}
-                placeholder="we're actively trying to improve"
+                placeholder={PLACEHOLDER_PREFIX + ' ' + PLACEHOLDER_TEXTS[this.state.category]}
```
vanderhoop: 2 things here. 
1. just inline the string `"Ex. " rather than declaring a whole new constant up above
1. use string interpolation with template literals. pretty sure the linter should barf with the concatenation

pulsedemon:
@vanderhoop We thought it would be better to have a separate constant for the prefix to make it easier to change in the future if we decided to do that.  Also, it might be cool to add more placeholders later on and randomize them to make it more interesting.  If we do that, the separate prefix would be nice.

vanderhoop:
i disagree on the prefix constant. it's only a three-character string, and we're only using it in one place, so right now the constant line is taking up space without providing real value.

pulsedemon:
@vanderhoop you're right.  It's gone

vanderhoop:
awesome. mergin'

Issue #113, opened 2017-04-24
send action items to all who have participated in a retro


vanderhoop:
As a facilitator,
I want to send action items to all those currently present in the retro,
So that the majority of those who have taken part receive the action items.

__By Way of Explanation:__ The initial action items email story hardcoded the recipient (to vanderhoop@me.com... which is me). That's not particularly useful. This story ensures that all those users who have participated in a retro receive the action items email. 

__Pre-Reqs:__
- [x] [Persist user participation in a retro](https://github.com/stride-nyc/remote_retro/issues/112)

__Acceptance Criteria:__
- [ ] all users who have participated in a retro receive the action items email when the facilitator sends it out.

__Helpful Links:__
- [BambooTest](https://hexdocs.pm/bamboo/Bamboo.Test.html)
  - provides some really helpful assertions for testing sending emails through Bamboo.

vanderhoop:
implemented

PR #120, opened 2017-04-27
Send action items to Retro participants

Zanadar:
Addresses https://github.com/stride-nyc/remote_retro/issues/113


web/channels/retro_channel.ex
```diff
@@ -25,10 +25,9 @@ defmodule RemoteRetro.RetroChannel do
   end
 
   def handle_in("proceed_to_next_stage", %{"stage" => "action-item-distribution"}, socket) do
-    email_send_status = Emails.action_items_email(socket.assigns.retro_id)
-                        |> Mailer.deliver_now
+    email_send_status = Emails.action_items_email(socket.assigns.retro_id) |> Mailer.deliver_now
```
Zanadar: yer dumb

test/channels/retro_channel_test.exs
```diff
@@ -46,13 +46,6 @@ defmodule RemoteRetro.RetroChannelTest do
 
   describe "pushing `proceed_to_next_stage` with a stage of 'action-item-distribution'" do
     setup [:join_the_retro_channel]
-
```
vanderhoop: why are we deleting this test? let's augment the expectation such that it expects it to be sent to the dev user's email.

web/models/user.ex
```diff
@@ -35,4 +35,14 @@ defmodule RemoteRetro.User do
     |> unique_constraint(:email)
     |> validate_format(:email, ~r/@/)
   end
+
```
vanderhoop: love this change, but confused that it's in *this* particular PR?

vanderhoop: would also be nice to get some tests around this, as it's part of `User`'s public interface

lib/remote_retro/emails.ex
```diff
@@ -1,13 +1,23 @@
...
+  @moduledoc false
   import Bamboo.Email, except: [from: 2]
   import Ecto.Query
   alias RemoteRetro.Repo
+  alias RemoteRetro.User
 
   def action_items_email(retro_id) do
     action_items = retro_action_items(retro_id)
 
+    participant_emails = Repo.all(
```
vanderhoop: this is great, but because there's logic in here, needs some tests

.credo.exs
```diff
@@ -74,7 +74,7 @@
 
         {Credo.Check.Readability.FunctionNames},
         {Credo.Check.Readability.LargeNumbers},
-        {Credo.Check.Readability.MaxLineLength, priority: :low, max_length: 80},
+        {Credo.Check.Readability.MaxLineLength, priority: :low, max_length: 120},
```
vanderhoop: i'm not against upping it, but 120 is a bit high. let's make it a 100?

web/models/user.ex
pulsedemon: @vanderhoop it's a part of this pr because we didn't want to duplicate that code in a test we were working on.  We ended up not needing the test.

test/channels/retro_channel_test.exs
pulsedemon: @vanderhoop We already have `SendActionItemToUsersViaEmailTest`, so we felt like this one was unnecessary.

lib/remote_retro/emails.ex
pulsedemon: @vanderhoop isn't this covered by `SendActionItemToUsersViaEmailTest`?

test/channels/retro_channel_test.exs
vanderhoop: Still necessary. This is a unit test that covers a particular action, while `SendActionItemToUsersViaEmailTest` is a system (feature) test. It may feel redundant, but this test can become more fine-grained over time, while the system test is a more general test of the overall happy-path for the feature.

web/models/user.ex
vanderhoop: nice makes sense. still a good change.

Zanadar: we  could/ should make it private though, until such time as its needed by other modules.

test/channels/retro_channel_test.exs
Zanadar: In this case, the only thing that matters for what this code does is the state of the socket (in fact the retro id). We're not using any of the data we push down the channel for anything. If this were truly a unit test, the code we were testing would have an output that we would be be asserting against. In Ruby I'd be reaching for a mock here, and while I know that Bamboo provides some kind of mock library honestly I think that its overkill.

If you're open to it, I'd like to come up with some some kind of testing based on the ideas Jose Valim expresses here: http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/ Otherwise, I'm not really sure how to test this without establishing a number of external dependencies, which kinda says to me that this code too coupled. I think we should fight against the Railsy tendency to just shove tons of records into a database and pretend that we're doing unit testing. Its slow doesn't really isolate the code we're testing and Elixir gives up tools to do better.

web/models/user.ex
vanderhoop: @Zanadar but you're calling it publicly, no?

test/channels/retro_channel_test.exs
vanderhoop: You're right in the semantic sense that this isn't a true unit test, and you're right that adding a bunch of test records would be a long walk and a bad habit to get into, but to delete the test outright is not the answer. It was small, fast, and covered a crucial side effect of this handler. Just maintain the current expectation of the `subject` field and we're golden.

pulsedemon: @vanderhoop The main reason we deleted the test was because the modifications to emails.exs broke the test and it would have been a lot of work to get the test passing again with little to no benefit

vanderhoop: just pulled this branch to do the rebase, and @Zanadar, *now* i get why you're talking about this test needing inserting test records. the test was failing because there was no-one in the `to` field of the email, which in turn was caused by there being no one in the test database. my bad.

vanderhoop: wasn't clear to me why you were talking about database inserts for this test

PR #126, opened 2017-04-28
updated preflight and readme with information about oauth setup

pulsedemon:


vanderhoop:
@pulsedemon This is great. Thanks for being thorough! mergin'!

PR #147, opened 2017-05-17
Clarified a few steps in the READMD file

roh:


vanderhoop:
this is great, @roh. thanks for taking it on!

PR #148, opened 2017-05-19
Fix #136 - use Repo.insert_or_update/1

### LINKED ISSUE #136: use `Repo.insert_or_update/2` to avoid conditional
see auth controller. we're calling `Repo.insert!` when no user found and `Repo.update!` when one *is* found. in its place, we can use `Repo.insert_or_update/2`

__Helpful Link:__
- [`Repo.insert_or_update/2`](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert_or_update/2)

Trevoke:
I also tried to clean up the code a little bit; I did that in a separate commit so if you don't care about the code reorg it's easy to discard.

### DIFF
--- a/web/controllers/auth_controller.ex
+++ b/web/controllers/auth_controller.ex
@@ -9,20 +9,20 @@ defmodule RemoteRetro.AuthController do
 
   def callback(conn, %{"code" => code}) do
     user_info = Google.get_user_info!(code)
-    user = Repo.get_by(User, email: user_info["email"])
-
     user_params = User.build_user_from_oauth(user_info)
 
-    user = if !user do
-      changeset = User.changeset(%User{}, user_params)
-      Repo.insert!(changeset)
-    else
-      changeset = User.changeset(user, user_params)
-      Repo.update!(changeset)
-    end
+    {:ok, user} =
+      case Repo.get_by(User, email: user_info["email"]) do
+        nil -> %User{}
+        user_from_db -> user_from_db
+      end
+      |> User.changeset(user_params)
+      |> Repo.insert_or_update
 
-    user = Map.delete(user, :__meta__)
-    user = Map.delete(user, :__struct__)
+    user =
+      user
+      |> Map.delete(:__meta__)
+      |> Map.delete(:__struct__)
 
     conn = put_session(conn, :current_user, user)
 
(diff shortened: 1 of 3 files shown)


web/controllers/auth_controller.ex
```diff
@@ -1,35 +1,20 @@
...
-
+    user = OAuth.Google.retrieve_internal_user(code)
     conn = put_session(conn, :current_user, user)
 
     redirect conn, to: get_session(conn, "requested_endpoint") || "/"
   end
 
   defp authorize_url! do
-    Google.authorize_url!(scope: "email profile")
+    OAuth.Google.authorize_url!(scope: "email profile")
```
Trevoke: Yes, well, if we do that the code will become less explicit. We don't care about `Google`, we care about `Google OAuth`.

web/controllers/auth_controller.ex
```diff
@@ -1,35 +1,20 @@
...
-      Repo.insert!(changeset)
-    else
-      changeset = User.changeset(user, user_params)
-      Repo.update!(changeset)
-    end
-
-    user = Map.delete(user, :__meta__)
-    user = Map.delete(user, :__struct__)
-
+    user = OAuth.Google.retrieve_internal_user(code)
```
Trevoke: Yes, well, if we do that the code will become less explicit. We don't care about `Google`, we care about `Google OAuth`.

vanderhoop:
hey @Trevoke. cherry-picked the first commit, but going to leave off the second one. the auth controller can definitely be cleaned up/made more expressive, but the database calls didn't feel right housed in the OAuth module. Closing PR, but first commit already on master. Thanks!

Trevoke:
@vanderhoop did you close this because you're not interested in further conversation?

vanderhoop:
@Trevoke tl;dr happy to discuss further, but let's chat tonight in person rather than in a back and forth on PRs!

the long version: didn't want to assume you'd want to invest more time than you already had, given what you said about the code-reorg commit being a discardable bonus commit. My initial thought on cleaning the controller up would be to just extract the non-expressive bits out into private functions. for example, the deletion of the  of `__meta__` and `__struct__` keys is confusing inline. Want to take this on? If so, i'll happily reopen.

vanderhoop:
@Trevoke additionally, if you had other ideas for cleaning up the controller, put em here!

PR #154, opened 2017-05-29
Update user reducer + action creator to handle array of users/presences

caporta:
@vanderhoop 

Yo dude. small PR here to update user reducer + action creator to handle array of users/presences rather than one user at a time. This approach is in greater parity with the Presence data being passed from the Phoenix channel

### DIFF
--- a/test/actions/user_actions_test.js
+++ b/test/actions/user_actions_test.js
@@ -2,10 +2,10 @@ import { expect } from "chai"
 
 import * as actionCreators from "../../web/static/js/actions/user"
 
-describe("addUser", () => {
-  const user = { given_name: "Tiny Rick" }
+describe("addUsers", () => {
+  const users = [{ given_name: "Tiny Rick" }]
 
   it("should create an action to add user to users list", () => {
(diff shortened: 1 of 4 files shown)


ssomnoremac:
looks fine, so when a new user is added, you just receive an array of users with the new user appended? And likewise, when user exists, new array with user popped?

caporta:
@ssomnoremac when the `presence_state` of the Phoenix Channel changes, we'll receive an updated `presences` object with the current participants (and accompanying metadata) on the socket. Rather than pushing & popping, we simply assign the current `presences` object to local state. We'll leverage this same pattern but assign to the Redux store rather than local state. This means `store.user` may be better suited as an object than an array, but that depends on if we want to keep the metadata or not and i'll likely know more re: that in the next couple days

ssomnoremac:
@caporta I see, would it make sense to assign the object and also maintain an array of the ids? It adds logic to the reducer but generally removes logic from the UI.

vanderhoop:
@ssomnoremac ideally, we'll keep the data in one place, as a separation can make it hard to keep track of what data is where as the application grows.

caporta:
shoulder 👍 from @vanderhoop

PR #158, opened 2017-06-04
Replace unicode characters with svg images for happy/sad/confused - story# 143

tnewell5:
Hey Travis!

Let me know if this can be merged!
Thanks,

Tanya


web/static/js/components/category_column.jsx
```diff
@@ -5,9 +5,9 @@ import styles from "./css_modules/category_column.css"
 
 function CategoryColumn(props) {
   const categoryToEmoticonUnicodeMap = {
```
vanderhoop: hey @tnewell5! thanks for picking this one up. Unfortunately, this implementation (for the moment) borks the action item column, as replacing the `i` tag outright with an img tag doesn't account for the action item icon. Rather than have you add a messy conditional, I'm wondering if @imaginista can cook up an SVG for the action item column as well. That way, we can move wholesale to image tags.

ianmcnally:
might wanna make a sprite too 😁 https://css-tricks.com/svg-sprites-use-better-icon-fonts/

web/static/js/components/category_column.jsx
imaginista: Check it!

[rocket-ship.svg.zip](https://github.com/stride-nyc/remote_retro/files/1072892/rocket-ship.svg.zip)

vanderhoop:
yo ian! didn't know you could do sprites with svgs. thanks for the heads up!

vanderhoop:
hey @tnewell5 do you have time to update this PR in the coming week, using Brett's new action-item icon? Lemmeknow. if you're swamped I can update it

vanderhoopnbc:
@tnewell5 any word on this?

PR #161, opened 2017-06-07
Feature/add idea action and reducer

caporta:
@vanderhoop PR from @kimberlysuazo and myself! 👀  pls


vanderhoop:
hey @caporta and @kimberlysuazo, thanks for taking this on. ❤️ . 

need to update our ebert settings so that it reads our eslint config (right now it's reading ebert's default eslint config), because your indentation is correct. 

anyway, nice work! mergin'

PR #162, opened 2017-06-07
Convert stateless components to ES6 consts

samdec11:
Some stateless components were using the old `function` syntax. Here I've converted them into `const`s.


vanderhoop:
lovely. thanks for doing this, @samdec11 ❤️

PR #165, opened 2017-06-08
Redux migration/augment ideas reducer

vanderhoop:
hey @caporta @crq @samdec11 @kimberlysuazo, thought i'd open a tiny PR for augmenting our ideas action creator and reducer so that it can handle the initial setting of pre-existing ideas when joining an in-progress retro. I'm new to redux, and while I think I'm on point, I appreciate any feedback. thanks!

### DIFF
--- a/web/static/js/reducers/idea.js
+++ b/web/static/js/reducers/idea.js
@@ -1,7 +1,7 @@
 const idea = (state = [], action) => {
   switch (action.type) {
     case "SET_IDEAS":
-      return [...action.ideas]
+      return action.ideas
     case "ADD_IDEA":
       return [...state, action.idea]
     default:
--- a/web/static/js/reducers/idea.js
+++ b/web/static/js/reducers/idea.js
@@ -1,5 +1,7 @@
 const idea = (state = [], action) => {
   switch (action.type) {
+    case "SET_IDEAS":
+      return [...action.ideas]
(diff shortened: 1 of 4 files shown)


web/static/js/reducers/idea.js
```diff
@@ -1,5 +1,7 @@
 const idea = (state = [], action) => {
   switch (action.type) {
+    case "SET_IDEAS":
```
kimberlysuazo: @vanderhoop If the idea is to fetch all of the ideas, I think this looks good. Would this be the same as ```return action.ideas``` ?

web/static/js/reducers/idea.js
```diff
@@ -1,5 +1,7 @@
...
+      return [...action.ideas]
```
kimberlysuazo: @vanderhoop If the idea is to fetch all of the ideas, I think this looks good. Would this be the same as ```return action.ideas``` ?

caporta: @vanderhoop I like it! and 👍 to @kimberlysuazo's idea here. returning `action.ideas` outright seems the most straightforward approach, and will preserve immutability as it's a brand new array

vanderhoop: glaring oversight on my part. thanks for the spot, @kimberlysuazo!

PR #172, opened 2017-06-17
Kill RemoteRetro's local state in favor of redux store

vanderhoop:
This PR shifts the remainder of `RemoteRetro` component's local state (`stage` and `inserted_at`) to the redux store!

In doing this work, we introduce a new action, `SET_INITIAL_STATE`, that allowed two additional refactors: 
  1. make a single dispatch from the retroChannel's join callback
  1. delete the "SET_IDEAS" action and reducer case, as the case that called for it is now handled by `SET_INITIAL_STATE`, which handles setting `ideas`, `stage`, and `insertedAt` on the store

### DIFF
--- a/web/static/js/components/remote_retro.jsx
+++ b/web/static/js/components/remote_retro.jsx
@@ -22,7 +22,6 @@ export class RemoteRetro extends Component {
       .receive("ok", retroState => {
         actions.setInitialState(retroState)
         actions.updateStage(retroState.stage)
-        actions.setIdeas(retroState.ideas)
       })
       .receive("error", error => console.error(error))
 
--- a/web/static/js/components/remote_retro.jsx
+++ b/web/static/js/components/remote_retro.jsx
@@ -5,6 +5,7 @@ import { Presence } from "phoenix"
 
 import * as userActionCreators from "../actions/user"
 import * as ideaActionCreators from "../actions/idea"
+import * as retroActionCreators from "../actions/retro"
 import * as AppPropTypes from "../prop_types"
 import Room from "./room"
 import ShareRetroLinkModal from "./share_retro_link_modal"
@@ -12,9 +13,7 @@ import ShareRetroLinkModal from "./share_retro_link_modal"
 export class RemoteRetro extends Component {
   constructor(props) {
     super(props)
-    this.state = {
-      stage: "idea-generation",
-    }
+    this.state = {}
   }
 
   componentWillMount() {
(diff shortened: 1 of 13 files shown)


web/static/js/components/remote_retro.jsx
```diff
@@ -23,6 +22,7 @@ export class RemoteRetro extends Component {
     retroChannel.join()
       .receive("ok", retroState => {
         this.setState(retroState)
+        actions.updateStage(retroState.stage)
```
vanderhoopnbc: Here, it's too early to remove the `setState` invocation, as we're still relying on the retro's `inserted_at` timestamp.

web/static/js/reducers/stage.js
```diff
@@ -1,5 +1,7 @@
 const stage = (state = "idea-generation", action) => {
   switch (action.type) {
+    case "SET_INITIAL_STATE":
+      return action.initialState.stage
```
caporta: @travis is it redundant to `SET_INITIAL_STATE` if we're already providing the `state` param with a default argument of `idea-generation`?

vanderhoopnbc: hey @caporta, great question. Because a participant can join a retro when the retro has already advanced to the `action-items` stage, `idea-generation` is really just a placeholder until the `SET_INITIAL_STATE` action gets fired with the definitive `stage` value persisted server-side.

I looked into just leaving state with no default value and got the following error:

image

*But!* because we need to have a default value in place on application bootstrap, we can totally just have an empty string to not be misleading. 

You cool with an empty string? Lemmeknow!

### Detour that you didn't ask to be taken on: 
Your question indirectly gets into the order of operations of application bootstrap. Right now, the OOO is `createReducer() -> retroChannel.join -> actions.setInitialState`. This is a bit much, as redux actually allows us to pass initial state directly to `createReducer`, and a better approach would be to call `retroChannel.join() -> createReducer()`, allowing us to kill the `SET_INITIAL_STATE` action entirely. The only thing keeping us from this approach is the hairy problem we discussed last week, where the retroChannel would need *access* to the store to dispatch actions to it, and we're assuming the store isn't created at that time. Let's discuss this further on Tuesday, as the elegant solution is likely staring us in the face.

caporta: a) Empty string definitely works! b) and yeah, that's definitely an approach worth exploring this week 👍

vanderhoopnbc:
shoulder 👍 from @caporta

PR #178, opened 2017-06-28
Add disabled prop to stageProgressionButton based on checkForActionItems func

qlaire:


test/components/room_test.js
```diff
@@ -28,6 +28,37 @@ describe("Room component", () => {
...
+        const roomComponent = shallow(
+          <Room
+            {...defaultProps}
+            currentUser={facilitatorUser}
+            stage="action-items"
+          />
+        )
+        const stageProgressionButton = roomComponent.find(StageProgressionButton)
+        expect(stageProgressionButton).to.have.length(1)
+        expect(stageProgressionButton.prop("buttonDisabled")).to.be.true // eslint-disable-line
```
vanderhoop: why the need to disable this line @qlaire?

test/components/room_test.js
```diff
@@ -28,6 +28,37 @@ describe("Room component", () => {
...
+      it("renders a disabled <StageProgressionButton>", () => {
+        const roomComponent = shallow(
+          <Room
+            {...defaultProps}
+            currentUser={facilitatorUser}
+            stage="action-items"
+          />
+        )
+        const stageProgressionButton = roomComponent.find(StageProgressionButton)
+        expect(stageProgressionButton).to.have.length(1)
```
vanderhoop: I think we can remove the `length(1)` expectations from both of these tests, as the StageProgressionButton is rendered in both cases

web/static/js/components/room.jsx
```diff
@@ -18,6 +18,13 @@ const Room = props => {
   const progressionConfig = stageProgressionConfigs[stage]
   const showActionItem = stage !== "idea-generation"
 
+  function checkForActionItems() {
```
vanderhoop: hmmm... the naming here can be more expressive. this may not be the right name, but something along the lines of `noActionItemsSubmitted` would be more expressive of what the method is doing

web/static/js/components/room.jsx
```diff
@@ -18,6 +18,13 @@ const Room = props => {
...
+    if (showActionItem) {
```
vanderhoop: let's leverage the short circuit operator here, so it's something like, 

```es6
return showActionItem && !props.ideas.some(idea => idea.category === "action-item")
```

web/static/js/components/stage_progression_button.jsx
```diff
@@ -66,6 +66,7 @@ class StageProgressionButton extends Component {
         <button
           className="fluid ui right labeled teal icon button"
           onClick={this.handleStageProgressionButtonClick}
+          disabled={this.props.buttonDisabled}
```
vanderhoop: let's destructure `buttonDisabled` from props up above

vanderhoop:
hey @qlaire and @ddipanfilo, really nice work on this. there are some syntactic suggestions and a function name change to address (see comments), but after those are addressed, merge it!

test/components/room_test.js
qlaire: @vanderhoop because of this: https://github.com/eslint/eslint/issues/2102

I'll add the name of the rule to be more clear.

PR #183, opened 2017-07-07
Add transition to highlight; Change icon on highlight

samdec11:


vanderhoop:
Monsieur Moneybags strikes again. mergin'

PR #187, opened 2017-07-11
Refactor/persist userid on idea records

qlaire:


test/support/test_helpers.ex
```diff
@@ -1,15 +1,33 @@
...
   use Wallaby.DSL
-  alias RemoteRetro.Repo
+  alias RemoteRetro.{Repo, User}
 
   def persist_idea_for_retro(context) do
-    %{idea: idea, retro: retro} = context
-    idea = Map.put(idea, :retro_id, retro.id)
+    %{idea: idea, retro: retro, user: user} = context
+    user =
+      if Map.has_key?(user, :id) do
```
vanderhoop: Rather than use the id here, and falling back to email in cases where there isn't an id, could we just augment the mock user with an id?

test/support/test_helpers.ex
```diff
@@ -1,15 +1,33 @@
...
+    idea =
+      Map.put(idea, :retro_id, retro.id)
+      |> Map.put(:user_id, user.id)
     idea = Repo.insert!(idea)
 
     Map.put(context, :idea, idea)
   end
 
+  def persist_user_for_retro(context) do
+    %{user: user} = context
```
vanderhoop: Not a blocker for this PR, but Elixir allows you to destructure a parameter, which would make sense here, as no other elements of `context` are used here

test/support/test_helpers.ex
vanderhoop: As a follow-up, if we persist a user specifically for a test, we should have access to that user via the test;s context (per the `persist_user_for_retro` helper below). So I think we'll always have access to an id. If that turns out to be true--I'm often wrong--let's get rid of the conditional here.

test/support/test_helpers.ex
```diff
@@ -1,15 +1,33 @@
...
-    idea = Map.put(idea, :retro_id, retro.id)
+    %{idea: idea, retro: retro, user: user} = context
+    user =
+      if Map.has_key?(user, :id) do
+        user
+      else
+        Repo.get_by(User, email: user["email"])
+      end
+    idea =
+      Map.put(idea, :retro_id, retro.id)
```
vanderhoop: Rather than call `Map.put/3` twice, you could use [`Map.merge/2`](https://hexdocs.pm/elixir/Map.html#merge/2)

vanderhoop:
really great work here, @qlaire. address the test helpers (comments above) and this puppy is mergeable!

test/support/test_helpers.ex
qlaire: don't I still need the rest of the context for line 27?

test/support/test_helpers.ex
qlaire: The problem was that the user needed to be in the test database in order to not get a foreign key constraint error, so adding an ID to the mock user didn't work because the mock_user was not persisted in the test database for the unit tests. However it was in the DB for the e2e tests, so I figured for those tests we should just use the user that's already there instead of persisting another one. It does lead to this clunky code though. I'll change to persisting a new user in the database for the e2e tests.

test/support/test_helpers.ex
vanderhoop: boy is my face red

PR #192, opened 2017-07-12
Fix remaining linter errors

samdec11:
```
remote_retro/test/actions/retro_actions_test.js
  24:1  error  Too many blank lines at the end of file. Max of 1 allowed  no-multiple-empty-lines

remote_retro/test/components/remote_retro_test.js
  11:9  error  'wrapper' is assigned a value but never used  no-unused-vars

remote_retro/test/components/stage_progression_button_test.js
  138:7  error  Expected an assignment or function call and instead saw an expression  no-unused-expressions

remote_retro/web/static/js/app.js
  8:8  error  Using exported name 'RemoteRetro' as identifier for default export  import/no-named-as-default

remote_retro/web/static/js/components/category_column.jsx
  16:9  error  img elements must have an alt prop, either with meaningful text, or an empty string for decorative images  jsx-a11y/img-has-alt

remote_retro/web/static/js/components/remote_retro.jsx
  33:9   warning  Unexpected alert                             no-alert
  80:52  error    'stage' is missing in props validation       react/prop-types
  80:59  error    'insertedAt' is missing in props validation  react/prop-types

remote_retro/web/static/js/components/share_retro_link_modal.jsx
  65:20  error  Arrow function should not return assignment  no-return-assign

remote_retro/web/static/js/components/stage_progression_button.jsx
  83:3  error  propType "config" is not required, but has no corresponding defaultProp declaration  react/require-default-props

remote_retro/web/static/js/reducers/user.js
  5:5  error  Unexpected lexical declaration in case block  no-case-declarations
```

This PR fixes everything except for the warning about the JS alert in the remote_retro.jsx file. 

Things to note:
* The no-named-as-default ESLint rule does not play well with Redux's pattern of `connect`ing classes and exporting the connected class as default (https://github.com/benmosher/eslint-plugin-import/issues/544), so I've turned this rule off for all JS files.
* The no-unused-expressions rule doesn't play well with Mocha (https://github.com/eslint/eslint/issues/2102) so I've turned this rule off for test files.

### DIFF
--- a/web/static/js/components/category_column.jsx
+++ b/web/static/js/components/category_column.jsx
@@ -13,7 +13,7 @@ const CategoryColumn = props => {
   return (
     <section className={`${category} ${styles.index} column`}>
       <div className={` ${styles.columnHead} ui center aligned basic segment`}>
-        <img src={`/images/${category}.svg`} height={iconHeight} width={iconHeight} alt={`${category}`} />
+        <img src={`/images/${category}.svg`} height={iconHeight} width={iconHeight} alt={category} />
         <p><strong>{category}</strong></p>
       </div>
       <div className={`ui fitted divider ${styles.divider}`} />
--- a/web/static/js/components/category_column.jsx
+++ b/web/static/js/components/category_column.jsx
@@ -13,7 +13,7 @@ const CategoryColumn = props => {
   return (
     <section className={`${category} ${styles.index} column`}>
       <div className={` ${styles.columnHead} ui center aligned basic segment`}>
-        <img src={`/images/${category}.svg`} height={iconHeight} width={iconHeight} />
+        <img src={`/images/${category}.svg`} height={iconHeight} width={iconHeight} alt={`${category}`} />
         <p><strong>{category}</strong></p>
       </div>
       <div className={`ui fitted divider ${styles.divider}`} />
--- a/web/static/js/reducers/user.js
+++ b/web/static/js/reducers/user.js
@@ -2,9 +2,10 @@ const user = (state = [], action) => {
   switch (action.type) {
     case "SET_USERS":
       return action.users
-    case "UPDATE_USER":
+    case "UPDATE_USER": {
       const { userToken, newAttributes } = action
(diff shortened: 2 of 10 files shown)


web/static/js/reducers/user.js
```diff
@@ -2,9 +2,10 @@ const user = (state = [], action) => {
   switch (action.type) {
     case "SET_USERS":
       return action.users
-    case "UPDATE_USER":
+    case "UPDATE_USER": {
```
vanderhoop: oh, *that's* the preferred syntax? I wondered why it was barfing.

web/static/js/components/category_column.jsx
```diff
@@ -13,7 +13,7 @@ const CategoryColumn = props => {
   return (
     <section className={`${category} ${styles.index} column`}>
       <div className={` ${styles.columnHead} ui center aligned basic segment`}>
-        <img src={`/images/${category}.svg`} height={iconHeight} width={iconHeight} />
+        <img src={`/images/${category}.svg`} height={iconHeight} width={iconHeight} alt={`${category}`} />
```
vanderhoop: no need to use the backtics with interpolation here. just put `category` between curlys (curlies?)

web/static/js/reducers/user.js
vanderhoop: for the `alt` attribute, that is

samdec11: Yeah, when you're setting a variable inside of a switch case the linter complains if you don't have it within its own block because each level of the switch will have access to the variables of other cases.

web/static/js/components/category_column.jsx
samdec11: Good point

vanderhoop:
Yo @samdec11, thanks for doing this, as it needed doing. @crq will sleep soundly tonight. the only blocker for merge is the interpolation referenced in [this comment](https://github.com/stride-nyc/remote_retro/pull/192#discussion_r127108116) above and you can merge

PR #201, opened 2017-07-20
Enhancement/pointing label prompt for idea

qlaire:
I ended up not checking if there were any ideas associated with the current user, because that check would've needed to be in a lifecycle hook like componentWillReceiveProps or componentDidUpdate, and it was causing an infinite loop of updating with the new ui reducer passing new props. So, instead I just show the pointer label when the current user hasn't entered anything, and take it away once they enter text in the input field.


vanderhoop:
hey @qlaire. looks much smaller than the original PR. nice work! going to create an issue for refactoring the IdeaSubmissionForm tests to not use setState, if possible. (I started that pattern, so it's me--not you). have a great weekend!

PR #202, opened 2017-07-21
Add prime directive component, and display when first joining retro

qlaire:


web/static/js/components/css_modules/prime_directive.css
```diff
@@ -0,0 +1,8 @@
+.wrapper {
```
vanderhoop: hey @qlaire, I think these styles can be avoided by using classes provided by semantic ui. as it provides means of centering content, limiting widths, et cetera. @tnewell5 can help you out, as we used semantic ui on the forecast manager project

web/static/js/components/room.jsx
```diff
@@ -1,20 +1,54 @@
...
-    <IdeaGenerationLowerThird {...props} />
-    <DoorChime {...props} />
-  </section>
-)
+const Room = props => {
+  if (props.stage === "prime-directive") {
+    const isFacilitator = props.currentUser.is_facilitator
+    const progressionConfig = stageProgressionConfigs[props.stage]
+    return (
+      <section className={styles.wrapper}>
```
vanderhoop: hmmm.... line 20 and 26 are identical. at the very least, we should avoid repeating that element. going a step further, it'd be nice to wrap the components for the prime directive stage in a `PrimeDirectiveStage` component, that simply wraps the `PrimeDirective`, the `UserList`, and (conditionally) the `StageProgressionButton`, while the any other stage would result in the render of the components in line 27-39. maybe i'm reaching, but it'd be nice to encapsulate the components needed in a particular stage in a particular component.

priv/repo/migrations/20170721171053_change_stage_default_value.exs
```diff
@@ -0,0 +1,9 @@
+defmodule RemoteRetro.Repo.Migrations.ChangeStageDefaultValue do
+  use Ecto.Migration
+
+  def change do
```
vanderhoop: hey @qlaire, can you change this migration so that there is an `up` and a `down` method? that way, we can roll this migration back to the previous default. the `change` method works in most migration cases, but here Ecto doesn't know what to roll the default back to

qlaire:
@vanderhoop thanks for reviewing! I've made the changes, and I added a new component for idea generation stage as well to keep consistent.

vanderhoop:
aww hellz yeah!

PR #214, opened 2017-07-26
Add branded header - story# 138

tnewell5:
Hey Travis!

Here's my first take at this story. My concern is all the css I've added that I was not able to nest. Let me know if you have refactoring and/or other suggestions on how to clean up my code.

Also, would it make sense to start using Less or Sass on Remote Retro?

Thanks,
Tanya


vanderhoop:
hey @tnewell5! several things:

- can you make a story for allowing nesting rules in our CSS modules? That'd be immensely helpful.
  - completion of said story won't be a blocker for this getting in.
- in cases where you're referencing colors (via hexidecimal), can you set and use css variables, as seen in [these lines](https://github.com/tnewell5/remote_retro/blob/219e2acedaae5ee1192414a4c99799491dea50a9/web/static/css/app.css#L2-L6)
- ensure the "Built by" text is centered with the stride logo, and in turn these should be aligned with
- remove the space between "Remote" and "Retro"
- ensure the "BETA" text doesn't sit distictly "above" the "RemoteRetro" text.

tnewell5:
hey Travis! Let me know if there is anything else that needs to be styled.

Thanks!
Tanya

vanderhoop:
mergin, cowfolk!

Issue #35, opened 2017-02-15
allow non-facilitators to delete a submitted idea within 5s of submission


williamcodes:
As a non-facilitator participant who has mistakenly added an idea to the board,
I would like to be able to delete that idea from the board,
So that I don't feel I have cluttered the board.

__Acceptance Criteria__
- [ ] non-facilitators see the delete button beside each of __their__ ideas
- [ ] after 5s have elapsed since the idea was persisted, the delete button fades out.
  - __Note:__ this was initially in story #53, which was written prior to the facilitator deletion story. The feeling is that we don't want to merge/deploy this deletion feature in an incomplete state, as ideas that have been on the board for 5s would allow folks to remove ideas that other participants have already taken in.
- [ ] when a non-facilitator deletes an idea they've submitted, all connected clients see the idea removed from their board.
- [ ] upon refreshing the page, the deleted item is not displayed

vanderhoop:
hey @williamcodes, love that you went ahead and made this issue. One small piece of feedback on the format: ideally, all user stories are written such that they reveal the value to the user. For example, this story could be written as 

> As a retro participant who has mistakenly added an idea to the board,
I would like to be able to delete that idea from the board,
So that I don't feel I have cluttered the board

And then the specifics of the UI interactions and results could be shifted down to an Acceptance Criteria  list where you can get more granular.

vanderhoop:
@williamcodes I realize you may have just fired this issue off to get it on the board, so sorry to nitpick in that case. I'm just happy you felt empowered to create it. Thanks!

williamcodes:
@vanderhoop that's great feedback, thanks, updating the story

Zanadar:
@vanderhoop Yo I wanna work on this one. Can I get added to the repo?

vanderhoop:
@Zanadar yo! if and when you make a pull request for this guy, can you make it to the `feature/deleting-submitted-ideas` branch? I want to keep this off master until story #39, "disallow users from deleting other users' ideas", is complete, as it could introduce a weird intermediate state that could be off-putting to users. Thanks!

PR #216, opened 2017-07-27
Feature/user can delete idea

qlaire:
__Description of Change(s) Introduced:__ a non-facilitator user can now delete their own idea, but not after 5 seconds of the idea being on the board

- ~~[moment.js](https://momentjs.com/) - 2.18.1~~
  - __for posterity:__ avoided the introduction of moment! [see comment below](https://github.com/stride-nyc/remote_retro/pull/216#issuecomment-318904493)

__Description of Testing Applied:__ tests for appearance of delete button, and setTimeout for getting rid of delete button

__Relevant github Issue:__ 

#35

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "classnames": "^2.2.5",
     "immutability-helper": "^2.1.2",
     "lodash": "^4.17.4",
+    "moment": "^2.18.1",
     "node-sass": "^3.8.0",
     "phoenix": "file:deps/phoenix",
     "phoenix_html": "file:deps/phoenix_html",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
     "classnames": "^2.2.5",
     "immutability-helper": "^2.1.2",
     "lodash": "^4.17.4",
-    "moment": "^2.18.1",
     "node-sass": "^3.8.0",
     "phoenix": "file:deps/phoenix",
     "phoenix_html": "file:deps/phoenix_html",
--- a/test/components/idea_controls_test.js
+++ b/test/components/idea_controls_test.js
@@ -103,6 +103,9 @@ describe("<IdeaControls />", () => {
     })
 
     context("when the user is not the facilitator", () => {
+      let clock
+      let earlierDate
+
       context("and the idea is not theirs", () => {
         it("doesn't render", () => {
           const retroChannel = { on: () => {}, push: sinon.spy() }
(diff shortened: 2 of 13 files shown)


vanderhoop:
Hey Claire, nice work. Two minor blockers to getting this in:

1) no need to bring in moment.js, as there's [a native js method for working with UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC). By leveraging it, we can put off adding a clientside dependency

2) we can avoid the complexity of the setTimeout and rerender by using raw css to make the delete button disappear after 5 seconds: https://stackoverflow.com/questions/21993661/css-auto-hide-elements-after-5-seconds. This has the added benefit of reusability across the app if we want to limit functionality to a window of time.

samdec11:
@vanderhoop Can you explain how to get from the ecto date string to a usable set of inputs for `Date.UTC`? That function's parameters are year, month, day, etc. and not a date string. We couldn't find a workable solution without Moment.

vanderhoop:
@samdec11 @qlaire I'd totally forgotten about this, but I faced a similar problem when only wanting to display the `ShareRetroLinkModal` to users who join within 5s of the retro's creation, and rather than do a bunch of JavaScript to make the value easily comparable, I changed the timestamp value further upstream, so that Phoenix would just give me a UTC datestring that I needed:

```elixir
  # web/models/retro.ex
  schema "retros" do
    # ...

    timestamps(type: :utc_datetime)
  end
```

This allowed me to pass the value in the way that the JavaScript `Date` constructor could digest in one go.
```es6
// web/static/js/components/share_retro_link_modal.jsx
const timeElapsedLessThanFiveSec = retroCreationTimestamp => {
  const millisecondsSinceRetroCreation = new Date(retroCreationTimestamp)
  const timeElapsedSinceRetroCreation = new Date().getTime() - millisecondsSinceRetroCreation
  return (timeElapsedSinceRetroCreation < 5000)
}
```

Note that if we made this change to the `Idea` model (joining the `Retro` model), it begs the question of whether *all* Phoenix models should just use `utc_datestring` timestamp for consistency.

qlaire:
@vanderhoop I got ride of the setTimeout and added the CSS animation. Thanks for the link!

I think it makes sense to be consistent across the the models and use utc_datetime, so I changed them all.

vanderhoop:
f*ck yes: mergin!

PR #219, opened 2017-07-31
add regex for required_version_num and use Regex.match

qlaire:
__Description of Change(s) Introduced:__ instead of hardcoding the version number for Hex, we should use a regex to check that `0.16.SOME_NUMBER` is being used


vanderhoop:
love it. thanks for taking this on, @qlaire. mergin'

PR #223, opened 2017-08-02
Add link to feedback survey to action item distribution alert

qlaire:
__Description of Change(s) Introduced:__ I created a Survey Monkey survey and added a link to it to the action item distribution alert. 

The survey results can be seen [here](https://www.surveymonkey.com/analyze/Sm5V2nM7ZHiVzOs34rhW81bWvjEWSpadfDicDUlGZF8_3D
).
username: remoteretro
pw: r3moter3tro8!

### DIFF
--- a/web/static/js/configs/stage_configs.jsx
+++ b/web/static/js/configs/stage_configs.jsx
@@ -1,3 +1,5 @@
+import React from "react"
+
 export default {
   "prime-directive": {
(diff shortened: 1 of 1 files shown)


vanderhoop:
woah, @qlaire! would never have thought to turn the configs file into jsx. very clever!

Issue #212, opened 2017-07-25
set up continuous deployment on CircleCI/Heroku


vanderhoop:
As a maintainer,
I would like to automate deployments to Heroku when CircleCI sees a new passing build on `master`,
So that we don't have to remember to deploy.

__By Way Of Explanation:__ as the pace of contributions has grown, I (and later @Zanadar and @samdec11) have become deployment bottlenecks. Let's eliminate that bottleneck asap to make our lives better.

__Acceptance Criteria:__

- [ ] passing builds of master on CircleCI result in deploys to our Heroku app, live at remoteretro.herokuapp.com

__Helpful Links:__

- [CircleCI - Continuous Deployment with Heroku](https://circleci.com/docs/1.0/continuous-deployment-with-heroku/)

PR #224, opened 2017-08-02
add continuous deployment config to circle.yml

qlaire:
__Description of Change(s) Introduced:__ I added continuous deployment config to circle.yml. I also followed the other steps [here]( https://circleci.com/docs/1.0/continuous-deployment-with-heroku/ ) so I think it should work!

__Relevant github Issue:__ 

#212


vanderhoop:
nice-icles. no way to see if it works without mergin', soooooo.... MERGIN!

PR #231, opened 2017-08-07
Add composite index to participations table - story#222

tnewell5:
Hey Travis!

Let me know if this is good to go. As per our discussion, I may not have enough rows for the index search to kick in (https://stackoverflow.com/questions/5203755/why-does-postgresql-perform-sequential-scan-on-indexed-column).

Thanks!
Tanya


vanderhoop:
perfecto! mergin'

Issue #235, opened 2017-08-08
persist users' votes 


vanderhoop:
As a retro participant,
I would like to reconnect to a retro I've left and see the correct vote count for all ideas on the board,
So that I can determine how to spend my votes using up-to-date information.

__By Way Of Explanation:__ in the interest of incrementally developing the voting feature, story #232 doesn't involve any sort of persistence. This story ensures that, upon reloading a retro, we see the correct vote counts for the ideas on the board.

__Acceptance Criteria:__
- [x] upon clicking the "Vote" button implemented in #232
  - [x] the vote_count is increased on the idea table
  - [x] the vote_count is increased on the participation table
  - [x] wrap the two SQL queries for doing the above in a transaction

- [x] upon reloading a retro with persisted votes, we see their vote counts on the board

__Helpful Links:__

- [`Ecto.Migration` docs](https://hexdocs.pm/ecto/Ecto.Migration.html)
- [`Ecto.Schema` docs](https://hexdocs.pm/phoenix/ecto.html#the-schema)

qlaire:
@vanderhoop just wondering why it's better to have a separate votes table instead of incrementing and decrementing a vote_count field on the user and idea. Lemme know!

vanderhoop:
hey @qlaire, i'm not against that approach. my original thinking was that votes wouldn't be anonymous, but based on the discussion in the #remote-retro channel last week, i'm all for users and vote

a couple things to keep in mind with this approach:
1. we'll want to increment the `vote_count` on the `participations` table rather than the `users` table, as a `participation` links a given user to the retro, and the vote limiting is scoped to a particular retro 
1. we don't want to increment an `idea`'s vote_count if the incrementing of the `participation`'s vote count somehow fails (and vice versa), so you'll want to wrap the two SQL queries in a transaction.

Issue #232, opened 2017-08-08
allow users to make unlimited votes on ideas in real time


vanderhoop:
As a facilitator,
I would like participants to collectively vote on the ideas that they feel are most valuable,
So that I can focus action-item generation/discussion _around_ those ideas.

__By Way Of Explanation:__ where in an in-person retrospective, we would allow users to group related ideas into categories before voting, we are developing the voting phase in the lowest cost way we feel feasible, by letting users vote on individual ideas. Category-based voting will come at a later date.

__Acceptance Criteria:__

- [x] when in the `idea-generation` stage, the StageProgressionButton's copy reads "Proceed to Voting"
- [x] upon clicking the "Proceed to Voting" button, thus entering the new "voting" stage":
  - [x] the StageProgressionButton reads "Proceed to Action Items"
  -  [x] all ideas on the board lose any existing idea action buttons, and gain a "Vote" button that contains the idea's current vote count, starting at zero.
     - __Notes:__
       - conveniently, SemanticUI has some existing markup/classes we can leverage. [example markup and image below](https://github.com/stride-nyc/remote_retro/issues/232#issuecomment-320822984) 
       - the precise styling of this button (padding, font-size, positioning within the idea list item) will come later. for this feature, the button just needs to exist _somewhere_ in the idea list item
- [x] when one participant clicks the "Vote" button for a given idea, all connected clients see that idea's vote count increment appropriately.

__Helpful Links:__

- [SemanticUI Labeled Buttons](https://semantic-ui.com/elements/button.html#labeled)

vanderhoop:
```es6
<div className="ui labeled button" tabindex="0">
  <div className="ui green button">
    Votes
  </div>
  <a className="ui basic green left pointing label">
    5
  </a>
</div>
```

PR #244, opened 2017-08-10
Feature/voting

tnewell5:
Hey Travis!

This PR covers two following voting feature stories: #232 and #235
Let Claire or I know if you have any questions.

Thanks!
Tanya


test/components/idea_controls_test.js
```diff
@@ -172,4 +183,63 @@ describe("<IdeaControls />", () => {
       })
     })
   })
+
+  describe("the vote button", () => {
+    context("when the stage is not idea-generation", () => {
```
vanderhoop: without looking at the implementation, i'd think that the only time we'd render the vote button is during the voting stage, right? does the category of the idea need to come into play, given that there won't be any action items under _after_ the voting stage is over?

test/components/vote_counter_test.js
```diff
@@ -0,0 +1,62 @@
...
+    category: "sad",
+    body: "redundant tests",
+    user_id: 1,
+    user: {
+      given_name: "Phil",
+    },
+    vote_count: 0,
+  }
+
+  it("renders a button that says Vote", () => {
```
vanderhoop: we can delete this test, as the "Vote" text is static

test/components/vote_counter_test.js
```diff
@@ -0,0 +1,62 @@
...
+    const voteCounter = shallow(
+      <VoteCounter
+        retroChannel={{}}
+        idea={idea}
+      />
+    )
+    const label = voteCounter.find("a")
+
+    expect(label).to.have.length(1)
+    expect(label.hasClass("pointing label")).to.be.true
```
vanderhoop: let's just have one expectation here, that the test of the label matches the idea's current vote count. and no need to specify that it's a pointing label

web/static/js/components/idea_controls.jsx
```diff
@@ -21,6 +22,11 @@ const IdeaControls = props => {
   const highlightTitle = isHighlighted ? "De-Highlight Idea for Participants" : "Announce Idea to Channel"
 
   function renderIcons() {
+    if (stage !== "idea-generation" && category !== "action-item") {
```
vanderhoop: per the comment above, the condition here should just be whether it's the voting stage

web/static/js/components/idea_generation_lower_third.jsx
```diff
@@ -12,10 +12,12 @@ const IdeaGenerationLowerThird = props => {
 
   const isFacilitator = currentUser.is_facilitator
   const stageConfig = stageConfigs[stage]
-  const showActionItem = stage !== "idea-generation"
+  const showActionItem = !["idea-generation", "voting"].includes(stage)
 
-  function wereActionItemsSubmitted() {
-    return showActionItem && !ideas.some(idea => idea.category === "action-item")
+  function progressionDisabled() {
```
vanderhoop: why was this change necessary? it doesn't seem related to voting, but I'm wrong a lot?

web/static/js/services/retro_channel.js
```diff
@@ -47,6 +47,10 @@ const applyListenerCallbacks = (retroChannel, store, actions) => {
     actions.deleteIdea(deletedIdea.id)
   })
 
+  retroChannel.on("vote_submitted", votedOnIdea => {
```
vanderhoop: let's add a test for this!

web/channels/retro_channel.ex
```diff
@@ -76,6 +76,18 @@ defmodule RemoteRetro.RetroChannel do
     {:noreply, socket}
   end
 
+  def handle_in("submit_vote", %{"id" => id}, socket) do
+    idea = Repo.get(Idea, id)
```
vanderhoop: based on the handler i see in the retro, where we're only relying on `id` and `vote_count` attributes, there shouldn't be any need to preload the user here, which will save us a database hit. 

The select then update queries can be consolidated into an update query using the following syntax:

```elixir
    query = from i in Idea, where: i.id == ^id
    {_row_count, [updated_idea]} = Repo.update_all(query, [inc: [vote_count: 1]], returning: true)
```

Not the prettiest thing in the world, but gets the job done in one go.

web/static/js/components/idea_generation_lower_third.jsx
vanderhoop: let's hold off on this for this story, so that adding tests for itdoesn't hold up this PR

web/static/js/components/idea_board.jsx
```diff
@@ -8,7 +8,7 @@ import styles from "./css_modules/idea_board.css"
 const IdeaBoard = props => {
   const { stage } = props
   const categories = ["happy", "sad", "confused"]
-  const showActionItem = stage !== "idea-generation"
+  const showActionItem = !["idea-generation", "voting"].includes(stage)
```
vanderhoop: to avoid the `!` here, we could just say that we only want to show the action items during particular stages, rather than list the stages that we don't want, i.e. `["action-items", "closed"].includes(stage)`.

In the long term we want to extract these conditionals and have composed elements like `IdeaGenerationIdeaboard` and `ActionItemIdeaBoard`, as these in-component conditionals are only going to get more complex as we add stages.

vanderhoop: also, once we've done away with the bang, let's update the tests for this condition (it only renders when the stage is either "action-items" or "action-item-distribution"

test/channels/retro_channel_test.exs
```diff
@@ -172,4 +172,29 @@ defmodule RemoteRetro.RetroChannelTest do
...
+    end
+
+
+    @tag user: @mock_user
+    @tag idea: %Idea{category: "sad", body: "doggone keeper"}
+    test "results in the idea being updated in the database", %{socket: socket, idea: idea} do
+      idea_id = idea.id
+      push(socket, "submit_vote", %{id: idea_id})
+
+      :timer.sleep(50)
```
vanderhoop: not ideal, but i recall having to do something similar, due to the db insertion being fired off in another process by the channel

test/components/vote_counter_test.js
tnewell5: done

test/components/vote_counter_test.js
tnewell5: done

web/static/js/components/idea_generation_lower_third.jsx
tnewell5: This was a bug fix implemented per Sean's suggestion. He was under the impression that the progression button should be disabled until an idea has been submitted. Should we create a bug fix PR for this?

web/static/js/services/retro_channel.js
tnewell5: will do

web/channels/retro_channel.ex
tnewell5: will look into this

web/static/js/components/idea_board.jsx
tnewell5: okay

test/channels/retro_channel_test.exs
tnewell5: noted

web/static/js/components/idea_generation_lower_third.jsx
vanderhoop: yes please!

test/components/idea_controls_test.js
tnewell5: agreed to leave alone for now

PR #246, opened 2017-08-11
Change condition for disabling the stage progression button

tnewell5:
__Description of Change(s) Introduced:__ Previously the stageProgressionButton was only disabled when there were no action items. This bugfix makes it so that it is also disabled when there are no ideas.


tnewell5:
lol Travis you're so fast. We were already working on them when you commented.

Issue #245, opened 2017-08-11
Disable voting buttons once in Action-Items stage


tnewell5:
As a facilitator,
I would like the voting buttons to be disabled once in Action-Items stage,
So that participants are no longer able to submit votes after the voting stage is closed.

__Acceptance Criteria:__

- [ ] vote buttons are disabled so that when a participant clicks on 'vote', the number of votes remains the same
- [ ] vote button looks disabled (maybe grey or a lighter color)

vanderhoop:
@qlaire @tnewell5 maybe you already thought of this, but wanted to call it out: the voting button should not be visible once the retro has been closed, following the logic you're implementing in #182 (currently in progress)

PR #252, opened 2017-08-14
Add prop and conditionally disable button when not in voting stage

qlaire:
__Description of Change(s) Introduced:__ using the classnames library, conditionally style the voting button so that it is clear that it is disabled when not in the voting stage. Also, actually disable the button. 

I couldn't get the cursor appearance to go back to default for the disabled button. Despite adding in the CSS for it, for some reason the div that wraps the button and label seem to ignore it. @vanderhoop do you have any ideas about what might be going on here?

__Relevant github Issue:__ 

#245

### DIFF
--- a/test/components/idea_controls_test.js
+++ b/test/components/idea_controls_test.js
@@ -10,6 +10,7 @@ describe("<IdeaControls />", () => {
   const mockUser = { id: 2, is_facilitator: true }
   const ideaGenerationStage = "idea-generation"
   const votingStage = "voting"
+  const actionItemsStage = "action-items"
 
   describe("on click of the removal icon", () => {
     it("pushes an `delete_idea` event to the retro channel, passing the given idea's id", () => {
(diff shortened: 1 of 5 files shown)


vanderhoop:
hey @qlaire, i think we can get most of what we want here by conditionally applying the "disabled" class to the button, as Semantic UI provides some prefab styling. for disabled buttons

image

qlaire:
@vanderhoop I should've mentioned that I tried that. I thought it looked too grey, and was a little hard to see. I'll add a screen shot and you can let me know what you think.

vanderhoop:
you're right that the disabled version looks faint, but it gives us the functionality we need with the tracest addition of code. We can up the opacity in a later story.

qlaire:
@vanderhoop done!

vanderhoop:
wooooot!

ianmcnally:
hi @vanderhoop and @qlaire! Just been creeping here, and thought about this article on disabled buttons I read recently: https://axesslab.com/disabled-buttons-suck/. tl;dr - disabled buttons aren't very accessible. Keep up the great work!

PR #259, opened 2017-08-18
Feature/link vote to user

qlaire:

Added a migration to add vote_count field to the participations table, and updated the corresponding model. I updated the submit_vote callback to use a transaction to update both the idea and the participation with the incremented vote count. 

The update_all method doesn't ever return an error, so instead I check if the updated row count is 0 for either idea or participation or both. I'm not completely sure how to handle these cases, but returning an error message seemed like a good idea. Do you know of a better way to handle this?

__Description of Testing Applied:__ Added tests for retro_channel.ex and updated the tests for VoteCounter component

__Relevant github Issue:__

- #235

### DIFF
--- a/priv/repo/migrations/20170816203847_add_vote_count_to_participation.exs
+++ b/priv/repo/migrations/20170816203847_add_vote_count_to_participation.exs
@@ -1,0 +1,9 @@
+defmodule RemoteRetro.Repo.Migrations.AddVoteCountToParticipation do
+  use Ecto.Migration
+
+  def change do
+    alter table(:participations) do
+      add :vote_count, :integer, default: 0, null: false
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


vanderhoop:
Hey @qlaire! Regarding your note up above, rather than have 3 separate messages to handle the various unhappy paths, you could simplify things by ditching the `case` statement completely and just pattern match using `=` with the happy-path pattern you already have (as that pattern ensures you've updated *both* database records). The benefit to that approach is that we the Elixir runtime will barf if either the `participation` record or `idea` record can't be found, as by this point of the retro, we should be able to rely on the `participation` record being persisted (as tested in the retro controller), and the appropriate idea id being passed by the RetroChannel JavaScript client (as tested in the RetroChannel test). 

If you'd rather not go with that approach, you could keep the logic you have, but encapsulate it in a service module. Please note thought that with this approach we  *won't* want to broadcast the `"vote_submitted"` event in all cases. If there is an error key, we'll want to let the submitting client know, [which is totally doable](https://github.com/phoenixframework/phoenix/blob/v1.2/web/static/js/phoenix.js#L29-L31)

qlaire:
@vanderhoop Thanks! The former makes sense, so Aaron and I went with that.

vanderhoop:
nice, mergin!

Issue #237, opened 2017-08-08
sort ideas based on the number of votes received (descending) upon completion of voting


vanderhoop:
As a facilitator,
I would like the idea lists to sort by vote count upon completion of the voting phase,
So that the priority order of ideas up for (discussion/action-item generation) is clear to myself and participants.

__By Way Of Explanation:__ note that we could do this sorting _during_ the voting phase, but this could (and likely, would) result in user 1 upvoting a given idea, causing that idea to shift just milliseconds before user 2 clicks the area that the idea vacated, causing them to misvote. 

__Acceptance Criteria:__

- [x] upon shifting into the "action-items" stage, the idea columns are sorted by most votes DESC.
- [x] _nice to have:_ the sort is animated (peep the link Helpful Link below)

__Helpful Links:__

- [React Flip Move library](https://github.com/joshwcomeau/react-flip-move)

PR #260, opened 2017-08-20
Feature/sort ideas by most votes

qlaire:
__Description of Change(s) Introduced:__ Sort by vote count instead of id when in action-item or action-item-distribution phase. Animate the sorting with the FlipMove library.

- FlipMove - ^2.9.14

__Description of Testing Applied:__ added test for category_column

__Relevant github Issue:__ 

- #237

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "react": "^15.4.1",
     "react-css-transition-replace": "^2.2.1",
     "react-dom": "^15.4.1",
+    "react-flip-move": "^2.9.14",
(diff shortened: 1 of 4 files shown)


vanderhoop:
hey @qlaire, super excited to get this in! just a headsup, had to fix an interesting bug caused by the action item column being dynamically added, as that b0rks the coordinates that flip move relies on for sortable element's starting position in the viewport. you can peep the change in commit 66c2b61. mergin!

Issue #236, opened 2017-08-08
limit users to 5 votes per retro


vanderhoop:
As a facilitator,
I would like to limit users to 5 votes per retro,
So that they cannot unduly influence what ideas are nominated for discussion/action-item generation.

__By Way Of Explanation:__ stories #232 and #235 impose no limits on the number of votes a given user can submit. They represent a fine start, but we can't allow voter fraud, because this is America.

__Acceptance Criteria:__

- [x] upon submitting their 5th vote in a particular retro, a given user's "Vote" buttons are disabled in that retro.

vanderhoop:
blocked by #235

PR #271, opened 2017-08-28
Feature/limit user to five votes

qlaire:
A user can now vote only 5 times per retro. 

__Dependencies Introduced or Upgraded:__ (if applicable): 
Upgraded enzyme and added prop-types to replace deprecated React.PropTypes and react-test-renderer to fix linting error

- (Dependency) - (current version number): 
prop-types 15.5.10, enzyme 2.9.1, and react-test-renderer 15.6.1

Frontend tests for reducer and ensuring button is disabled when user has voted 5 times.
Backend validation tests to ensure vote count is between 0 and 5

__Relevant github Issue:__ 

- #236


test/reducers/user_vote_counter_test.js
```diff
@@ -0,0 +1,103 @@
...
+describe("userVoteCounter reducer", () => {
+  describe("when there is an empty action", () => {
+    describe("when no new state is passed", () => {
+      it("should return an empty object", () => {
+        expect(userVoteCounter(undefined, {})).to.deep.equal({})
+      })
+    })
+  })
+
+  describe("when there is a 'ENGAGE_LUNAR_LANDING' action", () => {
```
vanderhoop: i think this `describe` can be deleted, as it's definitely an unrecognized action type, which is covered in the `describe` below

test/support/test_helpers.ex
```diff
@@ -30,6 +30,15 @@ defmodule RemoteRetro.TestHelpers do
     Map.put(context, :participation, participation)
   end
 
+  def persist_maxed_out_participation_for_retro(context) do
```
vanderhoop: looks like there's a lot of duplication here with the existing `persist_participation_for_retro` helper. let's kill this helper and extend the existing helper so that you can pass it a vote count

web/static/js/components/idea_controls.jsx
```diff
@@ -20,14 +21,15 @@ const IdeaControls = props => {
...
+  const canVote = userVoteCounter[currentUser.id] < 5
 
   function renderIcons() {
     if (stage !== "idea-generation" && category !== "action-item") {
       return (
         <VoteCounter
           retroChannel={retroChannel}
           idea={idea}
-          buttonDisabled={stage !== "voting"}
+          buttonDisabled={stage !== "voting" || !canVote}
```
vanderhoop: rather than a bang here, let's make the variable up above more expressive, as it's only used in this one place and it's being inverted.

web/static/js/services/retro_channel.js
```diff
@@ -47,8 +47,9 @@ const applyListenerCallbacks = (retroChannel, store, actions) => {
     actions.deleteIdea(deletedIdea.id)
   })
 
-  retroChannel.on("vote_submitted", votedOnIdea => {
-    actions.updateIdea(votedOnIdea.id, { vote_count: votedOnIdea.vote_count })
+  retroChannel.on("vote_submitted", data => {
+    actions.updateIdea(data.idea.id, { vote_count: data.idea.vote_count })
```
vanderhoop: Hmmm... not suggesting we change it now, but this dual action dispatch feels dirty, in the same way that making two database updates feels dirty to represent one action, in that our voting state is spread across two tables in the db and now two objects on the store. If this continues to propagate, I have a feeling we're going to regret not properly normalizing and introducing a `votes` table might come back to bite us.

vanderhoop: @qlaire @ams340 while you're on the bench, can you keep your noses primed for this dual state/action spell propagating throughout the app. @kpthomas pinging you as well, as i'm guessing you thought of this when you did the initial persistence work with Claire a couple weeks ago.

web/static/js/reducers/user_vote_counter.js
```diff
@@ -0,0 +1,22 @@
+const userVoteCounter = (state = {}, action) => {
+  switch (action.type) {
+    case "SET_INITIAL_STATE":
+      return action.initialState.participations.reduce(
```
vanderhoop: hmmm... curious about the decision to introduce an object to house everyone's vote count. feels like YAGNI in that, at least for now, we only need the current user's vote count.

test/channels/retro_channel_test.exs
```diff
@@ -207,4 +220,20 @@ defmodule RemoteRetro.RetroChannelTest do
       assert participation.vote_count == 1
     end
   end
+
+  describe "pushing a `submit_vote` event to the socket when the participation has 5 votes" do
+    setup [:persist_user_for_retro, :persist_idea_for_retro, :persist_maxed_out_participation_for_retro, :join_the_retro_channel]
+
+    @tag user: @mock_user
+    @tag idea: %Idea{category: "sad", body: "JavaScript"}
+    test "results in the broadcast of the voted on idea and updated participation to all connected clients", %{socket: socket, idea: idea, user: user, participation: participation} do
```
vanderhoop: the test language is a little off here. this seems to be testing that the vote count isn't incremented beyond 5. let's update it

test/components/idea_controls_test.js
```diff
@@ -236,6 +246,7 @@ describe("<IdeaControls />", () => {
             retroChannel={retroChannel}
             currentUser={currentUser}
             stage={ideaGenerationStage}
+            userVoteCounter={{}}
```
vanderhoop: ouch. curious if there was any discussion surrounding just housing the vote count on the user objects on state, rather than introducing an additional object on state?

web/channels/retro_channel.ex
```diff
@@ -79,16 +79,28 @@ defmodule RemoteRetro.RetroChannel do
...
-      |> Repo.transaction
+    idea_changeset =
+      Repo.get(Idea, idea_id)
+      |> Idea.changeset(%{vote_count: idea_vote_count + 1})
+
+    participation_changeset =
+      Repo.get_by(Participation, user_id: user_id, retro_id: retro_id)
+      |> Participation.changeset(%{vote_count: particip_vote_count + 1})
+
+   %{valid?: is_participation_valid} = participation_changeset
```
vanderhoop: rather than destructuring here, i think (think) it would be cleaner to just reference the the `valid?` property using dot notation in the conditional below, as that's the only place the `is_participation_valid` variable is leveraged

web/static/js/components/shadowed_scroll_container.jsx
```diff
@@ -1,22 +1,23 @@
-import React, { PropTypes } from "react"
-import styles from "./css_modules/shadowed_scroll_container.css"
```
vanderhoop: ahhhh. we actually can't get rid of this line. i know it showed up in the linter due to `styles` not being referenced, but the `import` statement ensures that the global styles therein are added to the stylesheet.

To fix the issue, we could A) disable the rule for this line, or B) just scope the styles to the `ShadowScrollContainer` component by stripping the `:global` off the `.overlay` styles in the stylesheet, and referencing `className={styles.overlay}` on line 5 below.

Actually thinking about it, the latter is definitely what we'll want to go with, as I can totally see a dev coming along and creating a global class of "overlay". by scoping it with css module hashes, we'll be protected from that scenario

test/reducers/user_vote_counter_test.js
qlaire: done!

test/support/test_helpers.ex
qlaire: done!

web/channels/retro_channel.ex
```diff
@@ -79,16 +79,28 @@ defmodule RemoteRetro.RetroChannel do
 
   def handle_in("submit_vote", %{"ideaId" => idea_id, "userId" => user_id}, socket) do
     retro_id = socket.assigns.retro_id
-    idea_query = from i in Idea, where: i.id == ^idea_id
-    participation_query = from p in Participation, where: p.user_id == ^user_id and p.retro_id == ^retro_id
+    idea_vote_count = Repo.get(Idea, idea_id).vote_count
```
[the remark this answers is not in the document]
qlaire: done!

web/static/js/services/retro_channel.js
qlaire: Yes will do!

web/static/js/reducers/user_vote_counter.js
vanderhoop: so true, forgot that the `currentUser` is derived. I think adding a `voteCount` to each user on the store is the most sensible approach, as we'll get `currentUser.voteCount` for free *due* to the aforementioned derivation.

vanderhoop: Regarding a database refactor, no need for that now. Wasn't proposing that we change it now. I just want us to be vigilant about the state leaking into different corners of the app. if anyone notices it, they should raise the red flag.

qlaire:
@vanderhoop this should be good to go. Let us know!

web/static/js/reducers/users.js
```diff
@@ -27,10 +27,14 @@ const users = (state = [], action) => {
...
       return assignFacilitatorToLongestTenured(withDeparturesRemoved)
     }
-    case "UPDATE_USER": {
+    case "UPDATE_PRESENCE": {
       const { userToken, newAttributes } = action
       return state.map(user => (user.token === userToken ? { ...user, ...newAttributes } : user))
     }
+    case "UPDATE_USER": {
+      const { userId, newAttributes } = action
+      return state.map(user => (user.id === userId ? { ...user, ...newAttributes } : user))
```
vanderhoop: beautiful. accounts for users with multiple presences. you two da bes

Issue #269, opened 2017-08-24
Display number of votes left for participant


qlaire:
As a participant,
I would like to see how many votes I have left,
So that I can better plan how I would like to spend my votes.

__Acceptance Criteria:__

- [x] the idea submission form (housed in the lower third) is replaced with displays "X Votes Left", where X is the number of votes the user has left, starting from 5.
- [x] as a user's votes are persisted, X decrements appropriately.
- [x] to avoid the "X votes left" line shifting position as the X value changes, have X on one line, and "Votes left" centered below it. 

__Example:__
```
                                            5
                                       Votes Left
```

vanderhoop:
closing as

PR #273, opened 2017-09-05
Feature/display number of votes left for participant

qlaire:
__Description of Change(s) Introduced:__ lower third is replaced with votes left message during voting stage

__Description of Testing Applied:__ front end tests for conditional rendering of votes left message

__Relevant github Issue:__ 

- #269

### DIFF
--- a/test/components/idea_controls_test.js
+++ b/test/components/idea_controls_test.js
@@ -4,6 +4,7 @@ import sinon from "sinon"
 
 import IdeaControls from "../../web/static/js/components/idea_controls"
 import VoteCounter from "../../web/static/js/components/vote_counter"
+import voteMax from "../../web/static/js/configs/retro_configs"
 
 describe("<IdeaControls />", () => {
   const idea = { id: 666, category: "sad", body: "redundant tests", user_id: 1 }
@@ -264,7 +265,7 @@ describe("<IdeaControls />", () => {
     context("when the currentUser has voted 5 times", () => {
       it("renders a disabled VoteCounter for the currentUser", () => {
         const retroChannel = { on: () => {}, push: sinon.spy() }
-        const currentUser = { id: 1, is_facilitator: false, vote_count: 5 }
+        const currentUser = { id: 1, is_facilitator: false, vote_count: voteMax }
 
         const wrapper = shallow(
           <IdeaControls
--- a/test/components/idea_controls_test.js
+++ b/test/components/idea_controls_test.js
@@ -4,7 +4,7 @@ import sinon from "sinon"
 
 import IdeaControls from "../../web/static/js/components/idea_controls"
 import VoteCounter from "../../web/static/js/components/vote_counter"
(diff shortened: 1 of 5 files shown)


test/components/idea_controls_test.js
```diff
@@ -4,6 +4,7 @@ import sinon from "sinon"
 
 import IdeaControls from "../../web/static/js/components/idea_controls"
 import VoteCounter from "../../web/static/js/components/vote_counter"
+import voteMax from "../../web/static/js/configs/retro_configs"
```
vanderhoop: this was confusing for me, i think due to the naming, because `voteMax` is the default export of a file that I would think exported an object. rather than having `retro_configs` have a default export of `voteMax`, it could just export the constant, and then here, we could just grab that via:
```es6
import { voteMax } from "../../web/static/js/configs/retro_configs"
```

test/components/idea_generation_lower_third_test.js
```diff
@@ -4,10 +4,14 @@ import { spy } from "sinon"
 
 import IdeaGenerationLowerThird from "../../web/static/js/components/idea_generation_lower_third"
 import StageProgressionButton from "../../web/static/js/components/stage_progression_button"
+import IdeaSubmissionForm from "../../web/static/js/components/idea_submission_form"
+
+import voteMax from "../../web/static/js/configs/retro_configs"
```
vanderhoop: same feedback as above!

test/components/idea_generation_lower_third_test.js
```diff
@@ -125,4 +129,66 @@ describe("IdeaGenerationLowerThird component", () => {
...
+          currentUser={userWithFiveVotes}
+        />
+      )
+
+      const votesLeft = lowerThird.find(".thirteen")
+
+      expect(votesLeft.text()).to.equal("0 Votes Left")
+    })
+
+    it("renders singular Vote if the user has one vote left", () => {
```
vanderhoop: Nice thinking here, but it the dynamic pluralization/singularization is a bit distracting in that the width of the text changes here. Let's kill this functionality and just have static text that says "Vote(s) Left"

web/static/js/components/idea_generation_lower_third.jsx
```diff
@@ -21,10 +22,26 @@ const IdeaGenerationLowerThird = props => {
     return noIdeasCreated || noActionItemsCreated
   }
 
+  function renderFormOrVoteCounter() {
+    if (stage === "voting") {
+      const userVoteCount = currentUser.vote_count
+      const votesLeft = userVoteCount ? voteMax - userVoteCount : voteMax
+      const votesText = votesLeft === 1 ? "Vote Left" : "Votes Left"
```
vanderhoop: see note above!

vanderhoop:
Hey @jarmstrng and @qlaire, i'm gonna take the changes requested above on, as it's content night and we'd like to get this in!

PR #303, opened 2017-11-14
Only display other users' vote counts after the voting phase - story# 277

tnewell5:
This PR introduces the following change:
- During the voting stage, users should only see their own vote count.
- After the voting stage, users should see each idea's total vote count.

Testing:
- Front-end unit tests were updated to reflect these changes.

### DIFF
--- a/web/static/js/components/vote_counter.jsx
+++ b/web/static/js/components/vote_counter.jsx
@@ -18,12 +18,19 @@ class VoteCounter extends React.Component {
   }
 
   render() {
-    const { buttonDisabled, votes, idea } = this.props
+    const { buttonDisabled, votes, idea, stage } = this.props
     const counterClasses = classNames("ui labeled right floated button", {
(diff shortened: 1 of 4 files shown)


web/static/js/components/vote_counter.jsx
```diff
@@ -40,7 +43,10 @@ class VoteCounter extends React.Component {
             transitionEnterTimeout={250}
             transitionLeaveTimeout={250}
           >
-            <div key={voteCountForIdea}>{voteCountForIdea}</div>
+            {stage === "action-items"
```
vanderhoop: @tnewell5 Rather than repeating the `<div>` (and, in turn, the key and innerValue) with different variables, let's just have a single variable, `voteCountForIdea` that we interpolate _into_ the `key` and `innerValue`. The naming of the variable will be clear enough for anyone who reads the logic, as it's a single `if (stage === 'voting') {} else {}`

PR #304, opened 2017-11-15
Choose less jarring entry and exit sounds - story# 253

tnewell5:
This PR replaces the existing chime with two chimes: one for joining and one for leaving the retro.
These sounds are supposed to be less annoying to participants.

No new dependencies introduced.

Front-end unit tests were added to cover the new chimes.

If you prefer different sounds, feel free to choose from the list:
https://www.zapsplat.com/sound-effect-categories/


vanderhoop:
@tnewell-nbc, these are great! Thanks for taking this on. Lot of negative feedback around the original chime sound :)

PR #315, opened 2017-11-21
Update mocha watcher configuration to compile jsx files on edit

damonblack:
The mocha test watch was previous ignoring any changes to the jsx files under test. This corrects the situation and makes TDD with the mocha watcher a happy time.

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,5 +1,6 @@
 test/**/*_test.js
 --color
+--compilers jsx:babel-register
 --require mock-css-modules
 --require babel-register
 --require=env-test


vanderhoop:
hey @dblack, this PR does the trick, but results in the following deprecation warning, as seen below:

image

I was able to get the desired behavior by invoking `mocha --watch --watch-extensions js,jsx`. To get the desired behavior without all the typing, we could create an npm script like `npm run tdd`. That would allow us a) get the desired behavior, b) roll with the local mocha, doing away with the global install, and c) avoid the deprecation warning introduced by the `compiler` directive. Thoughts?

vanderhoop:
closing in favor of an npm script, as discussed with @damonblack

Issue #309, opened 2017-11-20
Refactor: decouple `IdeaBoard` from default retro format


vanderhoop:
As outlined in #307, one of the pre-requisites of adding additional retro formats is to decouple components from the current retro format. 

The `IdeaBoard`, for example, will need to take a renderable list of categories, as opposed to the [static implementation currently in place](https://github.com/stride-nyc/remote_retro/blob/master/web/static/js/components/idea_board.jsx#L13), which *knows* the categories of the Happy/Sad/Confused format.

__Acceptance Criteria:__
- [ ] The `IdeaBoard` takes `categories` as props, rendering columns dynamically based on the given props.
  - __Note:__ For now, these props can be passed statically within the `IdeaGenerationStage` components, which renders the `IdeaBoard`.

PR #320, opened 2017-11-28
Refactor categories from the IdeaBoard

tnewell5:
This PR replaces hardcoded categories in IdeaBoard with categories passed in as props.
The categories have been placed into a separate constants file.

No new dependencies.
Tests have been updated accordingly.

Relevant github Issue:
- [309](https://github.com/stride-nyc/remote_retro/issues/309)

### DIFF
--- a/web/static/js/components/idea_board.jsx
+++ b/web/static/js/components/idea_board.jsx
@@ -9,12 +9,12 @@ import STAGES from "../configs/stages"
 const { ACTION_ITEMS, CLOSED } = STAGES
 
 const IdeaBoard = props => {
-  const { stage } = props
-  const categories = ["happy", "sad", "confused"]
+  const { stage, categories } = props
   const showActionItem = [ACTION_ITEMS, CLOSED].includes(stage)
-  if (showActionItem) { categories.push("action-item") }
+  const renderableColumnCategories = [...categories]
+  if (showActionItem) { renderableColumnCategories.push("action-item") }
 
-  const categoryColumns = categories.map(category => (
+  const categoryColumns = renderableColumnCategories.map(category => (
     <CategoryColumn {...props} category={category} key={category} />
   ))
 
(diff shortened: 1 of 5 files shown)


web/static/js/components/idea_board.jsx
```diff
@@ -9,10 +9,9 @@ import STAGES from "../configs/stages"
 const { ACTION_ITEMS, CLOSED } = STAGES
 
 const IdeaBoard = props => {
-  const { stage } = props
-  const categories = ["happy", "sad", "confused"]
+  const { stage, categories } = props
   const showActionItem = [ACTION_ITEMS, CLOSED].includes(stage)
-  if (showActionItem) { categories.push("action-item") }
+  if (showActionItem && !categories.includes("action-item")) { categories.push("action-item") }
```
vanderhoop: Hey @tnewell5. Rather than adding a check to see if we've already pushed the new category into the collection. let's just create a constant like `renderableColumnCategories = [...categories, "action-item"]`. Reason being, we don't want to mutate a constant, which `Array.prototype.push` does.

web/static/js/components/remote_retro.jsx
```diff
@@ -6,6 +6,7 @@ import * as AppPropTypes from "../prop_types"
 import Room from "./room"
 import Alert from "./alert"
 import DoorChime from "./door_chime"
+import CATEGORIES from "../configs/categories"
```
vanderhoop: @tnewell5, rather than house this in its own file, let's juse export `categories` as a constant from the `retro_configs.js` file

web/static/js/components/idea_board.jsx
tnewell5: cool - thanks!

web/static/js/components/remote_retro.jsx
tnewell5: will do!

Issue #333, opened 2017-12-05
bugfix: avoid overlap by having vote counts and stage progression button sit inline


vanderhoop:
As a user in the voting stage,
I want the lower third to not sit atop the bottom of the idea columns
So that i can read the ideas at the bottom of columns in an unobstructed way.

__By Way Of Explanation:__ when we implemented the voting lower third, we added vertical space by making the stage progression button sit _below_ the votes left indicator. this causes the following problem:

![overlap](https://user-images.githubusercontent.com/4386645/33616285-b23aae26-d999-11e7-8051-576ef55daa00.png)

__Acceptance Criteria:__

In the voting stage:
  - when the user is the facilitator:
    - [ ] the stage progression button sits to the right of the "3\nVotes Left" indicator, rather than below it.
      - __Note:__ while this change will save vertical space, it may cause the "3\nVotes Left" indicator to not be centered on the screen (which is the current design). That's fine.
    - [ ] when the sad column is full, the heads don't overlap the ideas
  - when the user is a non-facilitator:
    - [ ] the "3\nVotes Left" indicator appears centered on the screen (the current design)

__Helpful Links:__

- [SemanticUI Grid Docs](https://semantic-ui.com/collections/grid.html)

PR #338, opened 2017-12-11
Bugfix/avoid overlap in voting phase - story# 333

tnewell5:
This PR introduces a bug fix for the overlap of participants and the ideas during the voting phase.

__Dependencies Introduced or Upgraded:__ n/a

__Description of Testing Applied:__ tests were updated for the conditional rendering

__Relevant Screenshots/GIFs:__ 
facilitator's view: 
image

non-facilitator's view:
image

__Relevant github Issue:__

- [333](https://github.com/stride-nyc/remote_retro/issues/333)

### DIFF
--- a/web/static/js/components/voting_lower_third_content.jsx
+++ b/web/static/js/components/voting_lower_third_content.jsx
@@ -7,8 +7,10 @@ import * as AppPropTypes from "../prop_types"
 
 const VotingLowerThirdContent = props => (
   <div className="ui stackable grid basic attached secondary center aligned segment">
-    <VotesLeft currentUser={props.currentUser} />
-    <div className="five wide computer eight wide tablet column">
+    <div className="two wide column">
+      <VotesLeft currentUser={props.currentUser} />
+    </div>
+    <div className="four wide column">
       <StageProgressionButton {...props} />
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


web/static/js/components/voting_lower_third_content.jsx
```diff
@@ -5,14 +5,28 @@ import StageProgressionButton from "./stage_progression_button"
...
+      </div>
+      <div className="four wide column">
+        <StageProgressionButton {...props} />
+      </div>
     </div>
-  </div>
-)
+  )
+
+  const nonFacilitorComponent = (
```
vanderhoop: hey @tnewell5, here we're repeating almost all of the markup for the component, where the only difference between facilitator and non-facilitator interfaces is the column containing the stage progression button. to keep this DRY, you could use the short-circuit operator, something like so:

```es6
<div className="ui stackable grid basic attached secondary center aligned segment">
  <div className="two wide column">
    <VotesLeft currentUser={props.currentUser} />
  </div>
  { props.currentUser.is_facilitator && 
    <div className="four wide column">
      <StageProgressionButton {...props} />
    </div>
   }
</div>
```

vanderhoop:
hey @tnewell5, other than the feedback about DRYing up the conditional render, can you set up the columns so that the stage progression button is on the right side of the lower third, like it is in the other stages?

web/static/js/components/voting_lower_third_content.jsx
tnewell5: done!

vanderhoop:
just a heads up, tanya. gonna merge this, but the empty column you added needs to also be scoped to the facilitator view, as it pushes the vote counts over for all other participants

image

PR #341, opened 2017-12-15
Add link to Google+ API to README

ianmcnally:

Updated README to include G+ API link. I totally missed this on my install.


vanderhoop:
@ianmcnally thanks for doing this! I'll merge tonight, as our CI/CD pipeline takes down the app on restart and a lot of folks conduct retros on fridays. On a separate note, how do you feel about the dependency of being online and signing into google during development?

ianmcnally:
dawg, always happy to pitch in.

it'd be great if we didn't have to be online. setting up the google api was a reason, albeit a little silly, that I hadn't set this up sooner. might be helpful to get some fixture(s) to speed up local dev, or something similar. in short, I like where your head is at.

PR #344, opened 2017-12-15
Accessible idea list

ianmcnally:

Following up on #343 (and branched off of), this removes the wrapping elements inside an idea list.

It also boosts accessibility of the idea generation page to 97 (screenshot below).

screen shot 2017-12-15 at 12 34 24 pm

### DIFF
--- a/web/static/js/components/idea_submission_form.jsx
+++ b/web/static/js/components/idea_submission_form.jsx
@@ -103,9 +103,9 @@ export class IdeaSubmissionForm extends Component {
           </div>
           <div className="eleven wide field">
             <div className="ui fluid action input">
-              <label htmlFor="idea-input-label" className="visually-hidden">Idea input</label>
+              <label htmlFor="idea-body-input" className="visually-hidden">Idea input</label>
               <input
-                id="idea-input-label"
+                id="idea-body-input"
                 type="text"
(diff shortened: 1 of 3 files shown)


vanderhoop:
@ianmcnally dude, nice catch on the react-flip-move version stuff. how'd you find that?

web/static/js/components/idea_submission_form.jsx
```diff
@@ -102,7 +103,9 @@ export class IdeaSubmissionForm extends Component {
           </div>
           <div className="eleven wide field">
             <div className="ui fluid action input">
+              <label htmlFor="idea-input-label" className="visually-hidden">Idea input</label>
```
vanderhoop: hey @ianmcnally, quick check here. can we swap this `id/htmlFor` combo from `'idea-input-label'` to `'idea-body-input'` to better reflect the markup. thanks!

ianmcnally:
@vanderhoop went to the docs to see how it worked and where i could move the flip component, then i found the little note about using fragments and 💡

web/static/js/components/idea_submission_form.jsx
ianmcnally: you got it. be871a2

Issue #248, opened 2017-08-11
Assignees for Action Items


qlaire:
As a participant,
I want to be able to assign action items to a particular user
So that that each action item has an accountable party.

__By Way Of Explanation:__ accountability is crucial for action items. if no one is named responsible for a given action item, the likelihood that it gets finished goes way down. To this point, the only way to "assign" an action item was to add an assignees name to the "body" of an idea. 

__Acceptance Criteria:__

- in the "action-items" stage, the lower third looks much like the `IdeaGenerationLowerThird`, but: 
  - [ ] "Category:" is replaced with "Assignee:"
  - [ ] In place of the category select dropdown, there is a different select dropdown populated with the full names of the users current present in the retro.
  - [ ] the form's "Submit" button is disabled until an assignee is selected
  - [ ] Once an action-item is submitted, any action item listed on the board has its assignee listed in parentheses after the idea's "body", but before the optional "(edited)" indicator
    - Example action item: `Take out the trash (Travis Bibimbap)`
      - when the idea's been edited `Take out the trash (Travis Bibimbap) (edited)`
  - [ ] upon reloading an idea board that has action items, its assignee is still listed either before or after (basically, the assignee needs to be persisted)

__Helpful Links:__

- [Action Item WikiPedia Page](https://en.wikipedia.org/wiki/Action_item)

williamcodes:
+1 from JOOR CTO

vanderhoop:
looks like @damonblack pushed up a branch for this: https://github.com/stride-nyc/remote_retro/tree/add_assignments

PR #348, opened 2017-12-27
Feature/assignees for action items

tnewell5:
This PR creates assignees for action items, visible in the 'action-item' column next to the action item. In order to create an action item, an assignee must be selected.

__Dependencies Introduced or Upgraded:__ n/a

__Description of Testing Applied:__ tests have been created/ updated as needed

__Relevant Screenshots/GIFs:__ 
image

__Relevant github Issue:__ 

- [issue 248](https://github.com/stride-nyc/remote_retro/issues/248)


web/channels/retro_channel.ex
```diff
@@ -35,6 +35,11 @@ defmodule RemoteRetro.RetroChannel do
     {:noreply, socket}
   end
 
+  def handle_in("user_typing_action_item", %{"userToken" => userToken}, socket) do
```
vanderhoop: hmmm... this is identical to the `"user_typing_idea"` handler defined up above. not sure we need both, because an action item *is* an idea.

web/channels/retro_channel.ex
```diff
@@ -60,6 +65,22 @@ defmodule RemoteRetro.RetroChannel do
     {:noreply, socket}
   end
 
+  def handle_in("new_action_item", %{"body" => body, "assigneeId" => assignee_id, "userId" => user_id}, socket) do
```
vanderhoop: Looks like we're repeating ourselves here, too, as the handler for `"new_idea"` does almost the exact same thing as this new handler, except it expects an `assigneeId` and hard-code's the category as `"action-item"`

web/static/js/components/action_item_generation_lower_third_content.jsx
```diff
@@ -0,0 +1,33 @@
...
+import * as AppPropTypes from "../prop_types"
+
+const ActionItemGenerationLowerThirdContent = props => {
+  const { stage } = props
+  const stageConfig = stageConfigs[stage]
+
+  return (
+    <div className="ui stackable grid basic attached secondary center aligned segment">
+      <div className="thirteen wide column">
+        <ActionItemSubmissionForm {...props} />
```
vanderhoop: __Note to self:__ this markup is nearly identical to the markup seen in the [IdeaGenerationLowerThirdContent](https://github.com/stride-nyc/remote_retro/blob/c003622b7aced10d20e886f605e6b9c8894b3f43/web/static/js/components/idea_generation_lower_third_content.jsx). The only difference is that it renders the action item submission form, which in turn is nearly identical to the regular idea submission form. I get that there are slight differences between the two forms (assignee selection, rather than category selection), but can we refactor the existing form elements, containing markup, and related action creators and handlers to be more reusable, rather than introducing a whole new set of nearly identical modules, components, et cetera?

web/static/js/services/retro_channel.js
```diff
@@ -60,6 +61,13 @@ const applyListenerCallbacks = (retroChannel, store, actions) => {
     })
   })
 
+  retroChannel.on("user_typing_action_item", ({ userToken }) => {
```
vanderhoop: see lines 57-62, as this handler's content is verbatim with the exception of the event name

vanderhoop: hmmm... we gotta do something about this duplication

web/static/js/reducers/ideas.js
```diff
@@ -10,6 +10,8 @@ const ideas = (state = [], action) => {
       ))
     case "DELETE_IDEA":
       return state.filter(idea => idea.id !== action.ideaId)
+    case "ADD_ACTION_ITEM":
```
vanderhoop: flagging as duplicate of lines 5-6

web/static/js/reducers/users.js
```diff
@@ -39,3 +39,8 @@ const users = (state = [], action) => {
 }
 
 export default users
+
+// SELECTORS
+export const getUser = (state, userId) => (
+  state.users ? state.users.find(user => user.id === userId) : {}
```
vanderhoop: is there a case where we wouldn't have users?

tnewell5: the selector was introduced by Damon. I only changed a part of it that was breaking tests (it used to return 'null' in the 'else' portion of the ternary but I changed it to '{}'.

vanderhoop: weird. i don't see an associated test for this selector. it's probably due to us testing the `connect`ed component (which will trigger the `mapStateToProps` call that uses this selector. Rather than change the implementation to accommodate an unrelated test, you could export the unconnected component and test that.

web/static/js/components/action_item_submission_form.jsx
```diff
@@ -0,0 +1,123 @@
+import React, { Component } from "react"
```
vanderhoop: flagging this component as a near duplicate of the idea generation form.

web/static/js/components/idea_submission_form.jsx
```diff
@@ -19,28 +18,24 @@ const PLACEHOLDER_TEXTS = {
 
 const pushUserTypingEventThrottled = throttle((retroChannel, currentUserToken) => {
   retroChannel.push("user_typing_idea", { userToken: currentUserToken })
-}, USER_TYPING_ANIMATION_DURATION - 100)
+}, USER_TYPING_ANIMATION_DURATION - 100, { leading: true })
```
vanderhoop: why the need for `leading: true` here?

samdec11: Damon added it in his copy of this file. I read the documentation for it and couldn't really argue for or against it, so I figured it was harmless? I can remove it if you'd like.

web/static/js/reducers/users.js
samdec11: We added a test for this function. Can you elaborate on what you meant by exporting the unconnected component to test that?

web/static/js/components/idea_submission_form.jsx
vanderhoop: yeah, let's remove it, as it's unrelated! and thanks for the quick response!

web/static/js/components/idea_submission_form.jsx
```diff
@@ -66,41 +62,62 @@ export class IdeaSubmissionForm extends Component {
...
+      pointingLabel = label("Submit an idea!")
+      fieldArgs = ["category", category, this.handleCategoryChange, defaultCategoryOptions]
+    } else if (!ideaEntryStarted && stage === ACTION_ITEMS) {
+      pointingLabel = label("Create Action Items!")
+      fieldArgs = ["assignee", assigneeId, this.handleAssigneeChange, [defaultOption, ...assigneeOptions]]
     }
 
+    const field = (labelName, value, changeHandler, options) =>
+      <div className={`${styles.flex} five wide inline field`}>
+        <label htmlFor={labelName}>{`${capitalize(labelName)}:`}</label>
```
vanderhoop: hmmm... biggest thing here is that the `label` and `field` consts could be extracted out into its own component, like `SelectDropdown` (or something like that). That way, we keep `IdeaSubmissionForm` free of this extra markup and logic, and we get the added benefit of (I think) being able to reuse that component in the future.

vanderhoop: note: the latter suggestion is maybe premature to think about, but I do feel that extraction of the `label` and `field` components, here defined inline, should be extracted out into their own wrapping module.

web/static/js/reducers/users.js
vanderhoop: for sure, but please note that I was only guessing at the cause of the issue Tanya mentioned above, so it could be wrong. Regarding exporting unconnected components, I think this doc sums it up better than I can: https://github.com/reactjs/redux/blob/master/docs/recipes/WritingTests.md#connected-components

For an example in the codebase, peep [Alert.jsx](https://github.com/stride-nyc/remote_retro/blob/master/web/static/js/components/alert.jsx) and [its associated test](https://github.com/stride-nyc/remote_retro/blob/master/test/components/alert_test.js)

web/channels/retro_channel.ex
```diff
@@ -45,6 +45,22 @@ defmodule RemoteRetro.RetroChannel do
     {:noreply, socket}
   end
 
+  def handle_in("new_idea", %{"body" => body, "category" => "action-item", "userId" => user_id, "assigneeId" => assignee_id}, socket) do
```
vanderhoop: hmmm... this handler is nearly identical to the handler defined below. I wonder if we can just delete it outright and pass a `null` `assigneeId` in cases where. The other alternative would be to extract the commonalities between this handler and the one above out into a private function, to avoid the repetition.

vanderhoop: lemmeknow whatcha think!

samdec11: Just pushed up a commit that does just that.

samdec11: Or, made one. Guess I haven't pushed it yet.

web/channels/retro_channel.ex
```diff
@@ -117,6 +109,29 @@ defmodule RemoteRetro.RetroChannel do
     {:noreply, socket}
   end
 
+  defp add_idea!(%{"body" => body, "category" => "action-item", "userId" => user_id, "assigneeId" => assignee_id}, socket) do
+    %Idea{
```
vanderhoop: hmmm... there's still a lot of duplication between these two definitions. The only difference seems to be the `category` hardcoding and the addition of an `assignee_id`. I'm curious if, from the the front, you passed `assignee_id` as `null`, you could just use just the one definition in all cases. The reason i suspect that might work is that the default `assignee_id` value in the db is `NULL` anyway. @tnewell5 Can you give that a shot?

web/static/js/components/idea_read_only_content.jsx
```diff
@@ -18,6 +19,8 @@ const IdeaReadOnlyContent = props => {
         stage={stage}
       />
       <span data-hj-masked>{ idea.body }</span>
+      { hasAssignee && <span className={styles.assignee}> ({assignee.given_name}</span> }
```
vanderhoop: no need for multiple spans here, as the user has a `name` attribute (provided for convenience by google in their oauth info) that is a concatenation of the user's first and last name.

web/static/js/components/idea_submission_form.jsx
```diff
@@ -87,20 +117,8 @@ export class IdeaSubmissionForm extends Component {
...
-              value={this.state.category}
-              className={`ui dropdown ${styles.select}`}
-              onChange={this.handleCategoryChange}
-            >
-              { this.props.showActionItem ? <option value="action-item">action-item</option> :
-                defaultCategoryOptions
-              }
-            </select>
-          </div>
+          {stage === IDEA_GENERATION && <SelectDropdown {...dropdownProps} />}
```
vanderhoop: rather than repeat the stage condition checks from up above (see lines 90 - 98), you could DRY this up by setting a let declaration to `let dropdownProps` and, here, saying `dropdownProps && <SelectDropdown {...dropdownProps} />`

tnewell5: done

test/components/idea_submission_form_test.js
```diff
@@ -15,24 +15,31 @@ describe("IdeaSubmissionForm component", () => {
     stopPropagation: () => undefined,
     preventDefault: () => undefined,
   }
+  const users = [
+    { id: 1, name: "Tina Fey" },
+    { id: 2, name: "Betty White" },
```
vanderhoop: Such excellent sitcom stars ❤️

test/components/idea_submission_form_test.js
```diff
@@ -165,13 +185,13 @@ describe("IdeaSubmissionForm component", () => {
...
             retroChannel={mockRetroChannel}
-            showActionItem={false}
             alert={{ herp: "derp" }}
+            users={users}
           />
         )
       })
 
-      it("passes the state's `category` to 'action-item'", () => {
+      it("passes the state's focus back to action item input", () => {
```
vanderhoop: nice catch 👍

test/reducers/users_reducer_test.js
```diff
@@ -149,4 +149,19 @@ describe("user reducer", () => {
       expect(newState).to.deep.equal([{ token: "abc123", name: "Tiny Rick", age: 70 }, { token: "zzz444", name: "Morty", age: 15 }])
     })
   })
+
+  describe("getUser", () => {
+    const state = {
+      users: [
+        { id: 1, name: "Betty White" },
+        { id: 2, name: "Rue McClanahan" },
+        { id: 3, name: "Estelle Getty" },
```
vanderhoop: https://media1.giphy.com/media/pKVGt2ypp5FYc/giphy.gif

web/static/js/components/idea_read_only_content.jsx
vanderhoop: 👀 !

vanderhoop:
hey @tnewell5! Nice work on this. Thanks for being such a good soldier with my feedback. There are a few small things I'd like to see taken care of, mentioned in the following comments, but I can also take care of these things this weekend, as I know you have client prep to take care of! Lemmeknow!

https://github.com/stride-nyc/remote_retro/pull/348#discussion_r159727352

https://github.com/stride-nyc/remote_retro/pull/348#pullrequestreview-86725553

https://github.com/stride-nyc/remote_retro/pull/348#discussion_r159727064

web/static/js/components/idea_read_only_content.jsx
tnewell5: done :)

web/channels/retro_channel.ex
tnewell5: about to tackle this bit - think it's the only one remaining

samdec11: @tnewell5 Since we're combining the two different cases into one, we don't actually need this `add_idea!` function. We can move the pattern matching and the Idea creation into the matching `handle_in` function clause.

Issue #350, opened 2017-12-29
enable the editing of an action item's assignee


vanderhoop:
As a participant who has mistakenly assigned an action-item to the wrong party,
I would like to be able to edit the action-item's assignee,
So that the correct party is accountable for the action-item.

__Acceptance Criteria:__

- [x] when editing an action item, the edit form:
  - contains a select dropdown with options for every user in the retro
  - the original assignee is pre-populated as the value of the select
- [x] upon changing the assignee and clicking "Save", the action item's attribution is updated across all connected clients, i.e. if you have an action item that reads "Deploy to gigalixir (Travis Vander Hoop)" and you change the assignee to someone named Tom Davey, the action item should update to "Deploy to gigalixir (Tom Davey)" across all connected clients

vanderhoop:
blocked by #248

PR #359, opened 2018-01-12
Edit action item assignee

samdec11:
__Description of Change(s) Introduced:__ Facilitators now have the ability to edit the assignee of an action item.

__Description of Testing Applied:__ Frontend component tests and backend to ensure expected messages are received and handled accordingly, changing the assigned assignee_id

screen shot 2018-01-12 at 10 58 04 am

- #350

### DIFF
--- a/test/components/idea_edit_form_test.js
+++ b/test/components/idea_edit_form_test.js
@@ -44,14 +44,14 @@ describe("<IdeaEditForm />", () => {
       }
 
       it("lists participants as potential assignees", () => {
-        const form = shallow(<IdeaEditForm {...testProps} />)
+        const form = mountWithConnectedSubcomponents(<IdeaEditForm {...testProps} />)
 
         const gripers = form.find("select[name='editable_assignee'] option")
         expect(gripers.map(option => option.text())).to.eql(["Helga Foggybottom", "Prudence Pumpernickel"])
       })
 
       it("shows the assigned user as selected initially", () => {
-        const form = shallow(<IdeaEditForm {...testProps} />)
+        const form = mountWithConnectedSubcomponents(<IdeaEditForm {...testProps} />)
 
         const value = form.find("select[name='editable_assignee']").props().value
         expect(value).to.equal(9)
@@ -159,7 +159,7 @@ describe("<IdeaEditForm />", () => {
           id: idea.id,
           body: idea.body,
           category: idea.category,
-          assignee_id: undefined,
+          assigneeId: undefined,
         })
       ).to.equal(true)
     })
--- a/test/components/idea_edit_form_test.js
+++ b/test/components/idea_edit_form_test.js
@@ -3,10 +3,13 @@ import { shallow } from "enzyme"
 import sinon from "sinon"
 
 import IdeaEditForm from "../../web/static/js/components/idea_edit_form"
+import STAGES from "../../web/static/js/configs/stages.js"
+
+const { IDEA_GENERATION, ACTION_ITEMS } = STAGES
 
 describe("<IdeaEditForm />", () => {
   const idea = { id: 999, body: "redundant tests", userId: 1 }
-  const stage = "idea-generation"
+  const stage = IDEA_GENERATION
   const currentUser = { is_facilitator: true }
   const mockRetroChannel = { on: () => {}, push: () => {} }
   const defaultProps = {
@@ -14,6 +17,13 @@ describe("<IdeaEditForm />", () => {
     currentUser,
     retroChannel: mockRetroChannel,
     stage,
+    users: [{
+      id: 7,
+      name: "Helga Foggybottom"
+    }, {
+      id: 9,
+      name: "Prudence Pumpernickel"
+    }],
   }
 
   describe("on initial render", () => {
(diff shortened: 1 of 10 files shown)


test/components/idea_edit_form_test.js
```diff
@@ -3,17 +3,27 @@ import { shallow } from "enzyme"
...
 
 import IdeaEditForm from "../../web/static/js/components/idea_edit_form"
+import STAGES from "../../web/static/js/configs/stages"
+
+const { IDEA_GENERATION, ACTION_ITEMS } = STAGES
 
 describe("<IdeaEditForm />", () => {
   const idea = { id: 999, body: "redundant tests", userId: 1 }
-  const stage = "idea-generation"
+  const stage = IDEA_GENERATION
```
vanderhoop: 💯  ❤️  💯

web/models/idea.ex
```diff
@@ -15,10 +15,11 @@ defmodule RemoteRetro.Idea do
   end
 
   @required_fields [:category, :body, :retro_id, :user_id]
+  @mutable_fields [:assignee_id | @required_fields]
 
   def changeset(struct, params \\ %{}) do
     struct
-    |> cast(params, @required_fields)
+    |> cast(params, @mutable_fields)
```
vanderhoop: beautiful

web/channels/retro_channel.ex
```diff
@@ -109,7 +109,7 @@ defmodule RemoteRetro.RetroChannel do
     {:noreply, socket}
   end
 
-  defp add_idea!(%{"body" => body, "category" => category, "userId" => user_id, "assigneeId" => assignee_id}, socket) do
+  defp add_idea!(%{"body" => body, "category" => category, "userId" => user_id, "assignee_id" => assignee_id}, socket) do
```
vanderhoop: why this change? not vehemently opposed to it, but it does seem inconsistent with the camelCased `userId` coming from the front end

web/static/js/components/idea_edit_form.jsx
```diff
@@ -9,7 +9,9 @@ class IdeaEditForm extends Component {
     this.state = {
       ideaBody: props.idea.body,
       ideaCategory: props.idea.category,
+      assignee_id: props.idea.assignee_id,
```
vanderhoop: here, we could keep the state properties consistent by adding the `idea` prefix

web/static/js/components/idea_edit_form.jsx
```diff
@@ -63,6 +71,16 @@ class IdeaEditForm extends Component {
             <option key={category} value={category}>{category}</option>
           ))}
         </select>}
+        {stage === "action-items" && <select
```
vanderhoop: a couple things here. The first is that we should attempt to re-use the `SelectDropdown` component that we introduced in the "action items assignment" story. The second is that we should leverage the stage constants found in the configs/stages.js file, both here (line 74) and [up above](https://github.com/stride-nyc/remote_retro/blob/96139549351ce2c9b93ced38eedf36c63e4ef683/web/static/js/components/idea_edit_form.jsx#L65)

web/static/js/components/idea_submission_form.jsx
```diff
@@ -49,7 +49,7 @@ export class IdeaSubmissionForm extends Component {
     const { currentUser } = this.props
     event.preventDefault()
     const newIdea = { ...this.state, userId: currentUser.id }
-    if (!newIdea.assigneeId) { newIdea.assigneeId = null }
```
vanderhoop: rather than doing this assignment just before submission, we can just set assigneeId to `null` where we declare `this.state = { ... }` in the constructor

web/static/js/components/idea_submission_form.jsx
```diff
@@ -65,12 +65,12 @@ export class IdeaSubmissionForm extends Component {
...
 
   handleAssigneeChange(event) {
-    this.setState({ assigneeId: Number.parseInt(event.target.value, 10) })
+    this.setState({ assignee_id: Number.parseInt(event.target.value, 10) })
   }
 
   render() {
     const { users, stage } = this.props
-    const { assigneeId, body, ideaEntryStarted, category } = this.state
+    const { assignee_id: assigneeId, body, ideaEntryStarted, category } = this.state
```
vanderhoop: the more i see it, the more the change to a snake_cased `assignee_id` (client-side, at least) seems to complicate things

web/static/js/components/idea_edit_form.jsx
vanderhoop: and in fact, we should also use the `SelectDropdown` [up above as well](https://github.com/stride-nyc/remote_retro/blob/96139549351ce2c9b93ced38eedf36c63e4ef683/web/static/js/components/idea_edit_form.jsx#L64-L73).

web/static/js/components/idea_submission_form.jsx
samdec11: Done

vanderhoop:
hey @samdec11, changes are looking good! Just need the `SelectDropdown` to take the pre-populated value. As right now, the "edit assignee" dropdown's initial value doesn't show anything. Fix that and we're golden!

screen shot 2018-01-14 at 9 44 19 am

web/static/js/components/idea_submission_form.jsx
```diff
@@ -101,7 +101,7 @@ export class IdeaSubmissionForm extends Component {
       pointerText = !ideaEntryStarted ? "Create Action Items!" : ""
       dropdownProps = {
         labelName: "assignee",
-        value: assigneeId,
+        value: assigneeId || "",
```
vanderhoop: maybe i'm being paranoid, but does this fallback value throw errors server side? i think we'll want assigneeId to explicitly be passed as `null` when passed to the server in cases where there is no assigneeId

samdec11: It doesn't send the fallback to the server. This is only there to stop console errors resulting from having the initial value from the state set as `null`.

PR #365, opened 2018-01-13
Serve local images from priv/assets

handlers:
If you use `copy-webpack-plugin` from `webpack-dev-server`, "copied" files will be served from `OUTPUT_PUBLIC_PATH`, as defined in the webpack's config. We need the files written disk because Phoenix is serving the assets (not webpack). So this PR adds `write-file-webpack-plugin`, which writes to disk. This plugin [has no effect](https://github.com/gajus/write-file-webpack-plugin) when `webpack` is used instead of `webpack-dev-server`, so I think we're all good if this config file is re-used.

See: [copy-webpack-plugin](https://github.com/webpack-contrib/copy-webpack-plugin#faq) for docs.

Start using `write-file-webpack-plugin` in `webpack.config.js`.

- write-file-webpack-plugin - ^4.2.0

None

Broken:
Broken local images

Fixed:
Fixed local images

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "redux": "^3.6.0",
     "semantic-ui-css": "2.2.10",
     "sinon": "^1.17.7",
-    "webpack": "^3.8.1"
+    "webpack": "^3.8.1",
+    "write-file-webpack-plugin": "^4.2.0"
   },
   "devDependencies": {
     "babel-core": "^6.17.0",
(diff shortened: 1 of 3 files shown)


vanderhoop:
@handlers! This is a really nice catch. This must have snuck by us when we migrated from webpack to webpack dev server, as many of the devs would have already had the assets written to disk. Thanks for the quick catch and fix!

Issue #191, opened 2017-07-12
Inform new facilitator of status & responsibility


caporta:
As a user who has just become a facilitator,
I would like be informed of my status & responsibilities,
so that I know what is expected of me.

__By Way Of Explanation:__
At the moment that the facilitator role passes from one user to another, responsibilities are implicitly granted, with no announcement of any kind. It would be better to explicitly lay out the responsibilities of the new facilitator.

__Acceptance Criteria:__

- [x] Recipient of facilitatorship should be alerted to their new status & responsibility via [the existing `Alert` interface](https://github.com/stride-nyc/remote_retro/blob/master/web/static/js/components/alert.jsx)
  - Alert's `headerText` reads: "Facilitation Transfer!"
  - Alert's body reads: "With {previousFacilitatorName}'s departure, you have assumed the role of facilitator. You'll be responsible for advancing the retro!"
- [x] other users are alerted to no such thing

caporta:
yo @imaginista, travis and I were talking about non-experienced facilitators (or just new users to the tool) being thrust into the facilitator role. How would they know what do do? We don't want to explain every little thing, but we were thinking it might make sense to at least let them know they'll be counted on to trigger stage progressions, edit vague ideas, et cetera. Is that a no-no from a UI perspective? How do we do that in such a way that it doesn't suck?

imaginista:
@caporta @vanderhoop 

Providing support and helpful information is critical to a good experience, but I don't think we should design some novel experience that instructs a user during an in-flight retro. My gut tells me we should design some cool little on-boarding one pager that someone can read 20 minutes before a RR and get the gist of the controls. 

Remember, Retrospectives are expensive. RR should be a tool that deserves the respect of the facilitator at hand to understand and feel confident before using. So, ideally, there are no in-flight facilitator transfers and if there are, the new facilitator could reference a help.remoteretro.com page for drop dead simple instructions on using it. 

IOW/TL;DR - I think we should design a help.remoteretro.com page.

imaginista:
@caporta @vanderhoop 

Riffing off above: im questioning the value of a facilitator-transfer ability. is this an ultra edge case? If someone has to bow out of leading a retro and the rest of the group wants to carry on, couldn't one of the participants just say "hey guys, I'm going to reschedule another retro to start in X minutes. Stand by for a new link"?

caporta:
@imaginista @vanderhoop  I think this feature has value primarily in two cases:

[...]

PR #373, opened 2018-02-01
Feature/alert for facilitator reassignment

qlaire:
__Description of Change(s) Introduced:__ Added new alert action for informing new facilitator of their responsibility to continue the retro. 

__Description of Testing Applied:__ test helper functions added for determining if the current user became the new facilitator and test alert action creator and reducer

__Relevant github Issue:__ #191

### DIFF
--- a/test/components/remote_retro_test.js
+++ b/test/components/remote_retro_test.js
@@ -1,7 +1,7 @@
 import React from "react"
 import { spy } from "sinon"
 
-import { RemoteRetro, isNewFacilitator, findCurrentUser, findFacilitatorName } from "../../web/static/js/components/remote_retro"
+import { RemoteRetro, isNewFacilitator } from "../../web/static/js/components/remote_retro"
 import STAGES from "../../web/static/js/configs/stages"
 
 const { IDEA_GENERATION, CLOSED } = STAGES
@@ -65,24 +65,3 @@ describe("isNewFacilitator", () => {
   })
 })
 
-describe("findCurrentUser", () => {
-  const stubUser = { token: "123" }
-  const stubPresences = [stubUser]
-  const stubUserToken = "123"
-
-  it("finds the user with the given token", () => {
-    expect(findCurrentUser(stubPresences, stubUserToken)).to.deep.equal(stubUser)
-  })
-})
-
-describe("findFacilitatorName", () => {
-  const stubFacilitator = {
-    is_facilitator: true,
-    name: "Jill",
-  }
-  const stubPresences = [stubFacilitator, { is_facilitator: false, name: "Bob" }]
-
-  it("finds the facilitator's name", () => {
-    expect(findFacilitatorName(stubPresences)).to.equal(stubFacilitator.name)
-  })
-})
--- a/test/components/remote_retro_test.js
+++ b/test/components/remote_retro_test.js
@@ -1,7 +1,7 @@
 import React from "react"
 import { spy } from "sinon"
 
-import { RemoteRetro } from "../../web/static/js/components/remote_retro"
+import { RemoteRetro, isNewFacilitator, findCurrentUser, findFacilitatorName } from "../../web/static/js/components/remote_retro"
 import STAGES from "../../web/static/js/configs/stages"
 
 const { IDEA_GENERATION, CLOSED } = STAGES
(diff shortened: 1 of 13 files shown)


test/reducers/alert_reducer_test.js
```diff
@@ -68,4 +69,27 @@ describe("alert reducer", () => {
       })
     })
   })
+
+  describe("when the action is CHANGE_FACILITATOR", () => {
+    const initialState = { headerText: "Warning!", bodyText: "You're being watched." }
```
vanderhoop: ❤️

test/components/remote_retro_test.js
```diff
@@ -1,7 +1,7 @@
 import React from "react"
 import { spy } from "sinon"
 
-import { RemoteRetro } from "../../web/static/js/components/remote_retro"
+import { RemoteRetro, isNewFacilitator, findCurrentUser, findFacilitatorName } from "../../web/static/js/components/remote_retro"
```
vanderhoop: There's a really nice opportunity to extract the `findCurrentUser`  and `findFacilitatorName` functions out into redux selectors, as we're already [determining the facilitator's name within the `LobbyStage` component](https://github.com/stride-nyc/remote_retro/blob/5515c41d36e68c19e49552cc99c5cd8102a7173c/web/static/js/components/lobby_stage.jsx#L12-L13), and by leveraging these selectors in the `mapStateToProps` call in RemoteRetro, we can greatly reduce the amount of logic that lives (or needs to be invoked) within `RemoteRetro#componentDidUpdate`. Allows us to DRY up the codebase a bit!

vanderhoop: A nice read [outlining and explaining the benefits of the redux selector pattern](https://hackernoon.com/selector-pattern-painless-redux-store-destructuring-bfc26b72b9ae)

laumontemayor: I've been looking into this, but there's a few issues. 

`findCurrentUser` is taking the `userToken` from the `window` object, so it's not coming from props or state.  I'm not sure it's going to be possible to make this a selector, since it seems you can only [create a selector with state or props (https://redux.js.org/docs/recipes/ComputingDerivedData.html).
 
As for `findFacilitatorName`, I tried to do it with the [Reselect library](https://github.com/reactjs/reselect), but it ended up being really complicated to test. I'm not sure it's worth making this a selector, since we're only using it once and the memoization doesn't seem necessary as it's going to be a frequent calculation.  

I do agree with you on the value of extracting these functions, especially `findCurrentUser`, so if there's any other way I can do this I'd be happy to try. Thoughts?

vanderhoop: hey @laumontemayor! All good points. Don't worry about `reselect`/memoization at this point, as that's an optimization pattern we can pursue down the road. You'll likely have a nicer time defining raw selection functions that we can use in `mapStateToProps`, as those will, at the very least, allow us to dedupe logic  that appears in multiple components.

Oh! And regarding `window`! It's available globally, so you can reference `window.userToken` wherever you define the selector. And if you need `window.userToken` defined in the selector's test, you can add `window.userToken = "someStringValue"` in the `before/beforeEach` block for the test. So we're covered both in-app and in the test suite.

vanderhoop: (my bad, didn't mean to submit that last comment yet) Regarding `findFacilitatorName`, we're deriving the facilitator name in `LobbyStage`, but also in `RemoteRetro`. By encapsulating this in a selector, we can dedupe that logic.

vanderhoop: If you have any other questions about this, hit me up here, but by all means you can also ping me on slack!

test/reducers/alert_reducer_test.js
laumontemayor: 😬

test/components/remote_retro_test.js
laumontemayor: See updated PR.

test/reducers/presences_reducer_test.js
```diff
@@ -153,3 +153,27 @@ describe("presences reducer", () => {
     })
   })
 })
+
+describe("selectors", () => {
+  describe("findCurrentUser", () => {
+    const stubUser = { token: "123" }
+    const stubPresences = [stubUser]
```
vanderhoop: hey @laumontemayor and @qlaire! not a blocker, but it would make the test more robust to have another presence in `stubPresences`, but with a different `token` value, as that communicates that value is the identifying criteria

web/static/js/components/lobby_stage.jsx
```diff
@@ -6,11 +6,10 @@ import StageProgressionButton from "./stage_progression_button"
 import ShareRetroLinkModal from "./share_retro_link_modal"
 import * as AppPropTypes from "../prop_types"
 import styles from "./css_modules/centered_text.css"
+import { findFacilitatorName } from "../reducers/presences"
 
 const LobbyStage = props => {
-  const { progressionConfig, currentUser, presences } = props
```
vanderhoop: ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 ❤️ 🔥 🎆 👍 💯 !!

vanderhoop: wait! looks like we aren't passing `facilitatorName` to the lobby stage!
image

Issue #376, opened 2018-02-05
disable facilitation transfer alerts in 'closed' stage


vanderhoop:
As a user who receives the facilitatorship during the 'closed' stage of a retro,
I do not want to be alerted to the transfer,
So that I am not bothered with useless information.

__By Way Of Explanation:__ this story is a small but crucial enhancement to the work outlined in #191, as transfers of facilitatorship (current determined by tenure in the retro) are most likely to happen in the 'closed' stage, due to the often rapid departures of users at the close of a retro. At the moment, there is the potential for each remaining user to be alerted to their new role as facilitator, even though there is no reasonable action can be taken during the 'closed' stage. We don't want to alert our users unnecessarily!

__Acceptance Criteria:__

- [x] when there is a facilitation transfer in the 'closed' stage, the user who would receive the facilitation transfer alert *doesn't* get alerted

vanderhoop:
(will remove the blocked label when #373 gets merged)

vanderhoop:
no longer blocked!

laumontemayor:
@vanderhoop PR for this: https://github.com/stride-nyc/remote_retro/pull/380

vanderhoop:
implemented! closing!

PR #380, opened 2018-02-14
disable alert for facilitation transfer on closed stage

laumontemayor:
* Disable alert and label of (Facilitator) for facilitation transfer when retro is in stage=closed

__Relevant github Issue:__ 

- [issue (376)](https://github.com/stride-nyc/remote_retro/issues/376)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ web/static/js/dll/*
 
 npm-debug.log*
 .eslintcache
+browser_logs.log
--- a/web/static/js/components/remote_retro.jsx
+++ b/web/static/js/components/remote_retro.jsx
@@ -8,6 +8,9 @@ import * as AppPropTypes from "../prop_types"
 import Room from "./room"
 import Alert from "./alert"
 import DoorChime from "./door_chime"
+import STAGES from "../configs/stages"
+
+const { CLOSED } = STAGES
 
 export function isNewFacilitator(prevCurrentUser, currentUser) {
   return ((prevCurrentUser.is_facilitator !== currentUser.is_facilitator)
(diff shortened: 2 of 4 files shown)


laumontemayor:
@vanderhoop Let me know if the tests are enough 😬

.gitignore
```diff
@@ -42,3 +42,4 @@ web/static/js/dll/*
 
 npm-debug.log*
 .eslintcache
+browser_logs.log
```
vanderhoop: ❤️ thanks for catching this! it didn't show up in my history, but i must have overlooked it!

web/static/js/components/remote_retro.jsx
```diff
@@ -26,7 +29,7 @@ export class RemoteRetro extends Component {
 
     if (prevProps.presences.length) {
       const prevCurrentUser = prevProps.currentUser
-      if (isNewFacilitator(prevCurrentUser, currentUser)) {
+      if (isNewFacilitator(prevCurrentUser, currentUser) && stage !== CLOSED) {
```
vanderhoop: hmmm... ideally we'd have unit tests for this condition, but the boilerplate necessary for triggering a `componentDidUpdate` that references props is lengthy and troublesome in and of itself. that said, it makes me think we should extract the facilitator transfer logic out of the component and into redux in the future. as it is all reliant on the `SYNC_PRESENCE_DIFF` action. The alert reducer could feasibly handle that, but it will need access to the all of the presences at that time, which it doesn't due to it's location in the state tree. i think we could solve that with redux-thunk, as that would allow us to augment the `SYNC_PRESENCE_DIFF` event with the current state of the presences.

That said, not a blocker. I'll look into that this weekend.

laumontemayor: Yeah, I was thinking about the same thing but decided that this was ok for now since the feature was important. I'll read a bit about redux-thunk and share my findings. 

🤘

vanderhoop: hey! peeped the `retroChannel` already has access to the `store`, and therefore, we could outfit the `PRESENCE_SYNC_DIFF` action with the `presences`, and have the alert reducer just handle that action. woot!

PR #381, opened 2018-02-14
Fix modal warning in console and tests

laumontemayor:
__Description of changes introduced:__
The following warning is shown in both the console and the e2e test output: 

` react-modal: App element is not defined. Please use `Modal.setAppElement(el)` or set `appElement={el}`. This is needed so screen readers don't see main content when modal is opened. It is not recommended `

This fix will set the element before the render with `componentWillMont()`

A few references: 
https://github.com/reactjs/react-modal/issues/133
https://stackoverflow.com/questions/48269381/warning-react-modal-app-element-is-not-defined-please-use-modal-setappeleme
https://github.com/reactjs/react-modal/issues/576

### DIFF
--- a/web/static/js/components/alert.jsx
+++ b/web/static/js/components/alert.jsx
@@ -8,12 +8,9 @@ import { actions as actionCreators } from "../redux"
 
 Modal.defaultStyles.content.zIndex = 2
 Modal.defaultStyles.overlay.zIndex = 2
+Modal.setAppElement("body")
 
 export class Alert extends Component {
-
-  componentWillMount() {
-    Modal.setAppElement("body")
-  }
 
   render() {
     if (!this.props.config) return null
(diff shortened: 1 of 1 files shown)


web/static/js/components/alert.jsx
```diff
@@ -10,6 +10,11 @@ Modal.defaultStyles.content.zIndex = 2
 Modal.defaultStyles.overlay.zIndex = 2
 
 export class Alert extends Component {
+
+  componentWillMount() {
```
vanderhoop: does this need to be set in `componentWillMount`? could we set it after the `zIndex` assignments on line 9 and 10?

vanderhoop: also, i think the app element might be selected via `.react-root`, rather than body, as that's the div that our react app mounts to

laumontemayor: Yes, we can set it after the `zIndex` and it works.

So I tried setting it to `.react-root` but the `alert-tests` fail, and I'm not entirely sure why. Example of one failure:

```
Alert component
       "before each" hook for "renders the headerText in a header text component":
     Error: react-modal: No elements were found for selector .remote-root.
```

vanderhoop:
woot!

Issue #349, opened 2017-12-29
deploy an instance of RemoteRetro to Gigalixir


vanderhoop:
As a developer,
I want to deploy RemoteRetro to [Gigalixir](https://gigalixir.com/),
So that we can fully leverage elixir's power.

__By Way Of Explanation:__ when we first developed RemoteRetro, the quickest and easiest way to get it deployed was via Heroku. Unfortunately, Heroku [has a handful of limitations when it comes to Elixir](https://hexdocs.pm/phoenix/heroku.html#limitations), including a connection limit for channels.

I have spoken with @jesseshieh, founder of Gigalixir, about the platform, and @erikreedstrom, a lead dev at [HomePolish](https://www.homepolish.com/), about his experience with Gigalixir, and I've come away feeling that Gigalixir is a strong fit for RR, as this will a) keep us from hitting Heroku's websocket connection limit, and b) allow us to leverage features of Elixir that are currently unavailable to us in production.

__Acceptance Criteria:__

- [x] instance of remote retro is deployed to Gigalixir

__Helpful Links:__

- [gigalixir.com](https://gigalixir.com/)
- [gigalixir docs](https://gigalixir.readthedocs.io/en/latest/)

jesseshieh:
Let me know if there's anything I can do to help!

vanderhoop:
hey @jesseshieh, not sure if you were serious about having time to open a PR with the necessary configs/boilerplate, but if you were, we'd certainly appreciate it!

jesseshieh:
No problem! I'll start on it tomorrow.

vanderhoop:
thanks, @jesseshieh!

jesseshieh:
I just submitted https://github.com/stride-nyc/remote_retro/pull/351, but it doesn't work yet. Just linking this issue with it.

vanderhoop:
closing, as implemented!

PR #382, opened 2018-02-15
Update App to use Secrets insted of Env Vars

nicholalexander:

This seems to fit the recommended pattern for secrets for Phoenix based on this documentation: https://hexdocs.pm/phoenix/deployment.html.

Update the way secrets are loaded in dev to use dev.secrets.exs.  This is imported in the config/dev.exs file.  

Add dev.secret.exs.example for easy copying.

Add dev.secret.exs to .gitignore.

Update readme to reflect new changes.

*Removed check_env_vars from Preflight*
This was no longer necessary, but we might want to have a preflight check in development that screens for the existence of the dev.secret.exs file and gives a similar helpful warning as was in the env var check method.

## Important:
*Change google client to use Application env instead of System Env.*
This would probably break production deployments as it is because any production deployments would have already set up the vars in the environment and not in the application.  If we want to make this change, we should consider how to address this!

Thanks!
nichol & laura

tagging: @laumontemayor


vanderhoop:
hey @nicholalexander and @laumontemayor! Thanks for looking into this. I'm against making this change at the moment, as we currently deploy RemoteRetro using Heroku, where secrets are passed via environment variables, (see https://hexdocs.pm/phoenix/heroku.html#making-our-project-ready-for-heroku), and I'd like to keep the means of application configuration consistent across environments, as different means of environment config would represent tech debt. But(!) due to some of the rate connection limits imposed by Heroku, we'll likely need to move off Heroku if and when RR sees larger traffic. See #349 for more info.

So while I don't think this change makes sense now, it may make sense when we migrate to Gigalixir.

vanderhoop:
Also: due to this being a possibility down the road, I'm going to leave the branch alive, as the work itself is solid.

nicholalexander:
@vanderhoop cool - makes sense!  i hadn't seen gigalixir, that also looks very cool.  happy to revisit this down the line and thanks for checking it out.

README.md
```diff
@@ -122,13 +122,20 @@ Next, click on "Credentials" in the left sidebar nav. On the right hand side, cl
 - Authorized JavaScript origins: `http://localhost:4000`
 - Authorized redirect URIs: `http://localhost:4000/auth/google/callback`
 
-Click on the Create button. Using the information Google provides, add the following lines to your profile and source (or open a new terminal).
+Copy the `dev.secret.exs.example` to `dev.secret.exs` and fill it out with the information Google provides. The final file should look like this:
```
vanderhoop: let's provide the bash `cp` command for these files, so the dev can just copy/paste the command

README.md
```diff
@@ -122,13 +122,20 @@ Next, click on "Credentials" in the left sidebar nav. On the right hand side, cl
...
+use Mix.Config
+
+config :remote_retro, :google_oauth,
+  client_id: "<Client Id>",
+  client_secret: "<Client Secret>",
+  redirect_uri: "http://localhost:4000/auth/google/callback"
+
 ```
 
+__Ensure that the new file is included in your `.gitignore`!__
```
vanderhoop: looks like you added the dev.secret.exs file to the gitignore, so we can kill this prompt!

lib/tasks/preflight.ex
```diff
@@ -3,27 +3,9 @@ defmodule Mix.Tasks.Preflight do
 
   @shortdoc "Checks for necessary ENVARS"
   def run(_) do
-    check_envars()
```
vanderhoop: I'm not sure about the evaluation order of mix tasks, but it could be that we can switch this to check for the Mix config defining non-empty values, rather than deleting the check outright. Can you check on this to see if it's a viable approach?

web/oauth/google.ex
```diff
@@ -28,9 +28,9 @@ defmodule RemoteRetro.OAuth.Google do
 
   defp client do
     @oauth_client.new([
-      client_id: System.get_env("REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_ID"),
-      client_secret: System.get_env("REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_SECRET"),
-      redirect_uri: System.get_env("REMOTE_RETRO_GOOGLE_OAUTH_REDIRECT_URI"),
+      client_id: Application.get_env(:remote_retro, :google_oauth)[:client_id],
```
vanderhoop: let's extract `Application.get_env(:remote_retro, :google_oauth)` out into a variable to avoid the repetition here

config/dev.secret.exs.example
```diff
@@ -0,0 +1,6 @@
+use Mix.Config
+
+config :remote_retro, :google_oauth,
+  client_id: "<Client Id>",
+  client_secret: "<Client Secret>",
+  redirect_uri: "http://localhost:4000/auth/google/callback"
```
vanderhoop: also, @nicholalexander! there are a few undesirable line-endings in this PR; might want to configure your editor such that those don't appear!

vanderhoop:
closing due to atrophy

PR #389, opened 2018-02-26
Use application aliasing in .iex.exs file for easier console work.

nicholalexander:
Automatically aliasing modules when working with the console (`iex -S mix`) makes it easier and allows you to not have to write RemoteRetro a thousand times.

You get to move from doing stuff like this:
`RemoteRetro.User.changeset(%RemoteRetro.User{}, u) |> RemoteRetro.Repo.insert!`

to doing stuff like this:
`User.changeset(%User{}, u) |> Repo.insert!`

This is implemented by having a .iex.exs file in the root and utilizing thoughtbot's Quick Alias library.  It is only in the dependencies for the dev environment as this is only focused on making it easier for devs to work with the console.

https://github.com/thoughtbot/quick_alias, 0.1.0.

Ran test suite.  No new tests added.


vanderhoop:
love it! mergin'

vanderhoop:
hmmmm, this breaks regular `iex` if you invoke it in your `remote_retro` directory, which is a pretty common use case if you want to play around in the repl. @nicholalexander can you look into that?

PR #390, opened 2018-02-26
Action item test

### LINKED ISSUE #366: end-to-end test for editing the assignee for an action item
hey @samdec11 and @qlaire! Nice work on implementing the action-item assignee. Merged the PR, and then realized after the fact that the PR didn't include the end-to-end test. 

__Acceptance Criteria:__
- [ ] end-to-end verification that the editing of an action-item's assignee is updated across clients

__Dev Notes:__
- if might make sense to bundle this verification into an existing end-to-end test for brevity-of-build reasons  reasons, but up to the implementer to make that call!

nicholalexander:
An e2e test which exercises reassigning an action item from an initial user to another user.  This uses three new test helpers, `persist_other_user_for_retro`, `assign_idea`, and `set_participation` to get the initial state correct.  It includes a first test to verify the action item is present and assigned to Other User (the non primary test user).  It then exercises the idea edit form and checks that it has been reassigned.

This also sets up a new `other_user` object in the test config.

It might make sense to clean up some of the helpers to generalize the `persist_user_for_retro` but wasn't sure how I'd do that - still figuring out the `context`.  Also, didn't update the tags on this test as I wasn't sure how you might want that set.

refactored with @laumontemayor

__Relevant github Issue:__
Resolves #366 
- [issue 366](https://github.com/stride-nyc/remote_retro/issues/366)

 Thanks!


test/support/test_helpers.ex
```diff
@@ -36,6 +37,27 @@ defmodule RemoteRetro.TestHelpers do
...
+    %{other_user: other_user} = context
+    user_params = User.build_user_from_oauth(other_user)
+    other_user =
+      User.changeset(%User{}, user_params)
+      |> Repo.insert!
+
+    Map.put(context, :other_user, other_user)
+  end
+
+  def assign_idea(%{other_user: other_user, retro: retro} = context) do
```
vanderhoop: hmmm... rather than hardcode an action items values here, i think we should alter the existing `persist_idea_for_retro` function such that it can be assigned to different users depending on the needs of the test. That way, we encapsulate idea creation in a single helper that can take dynamic category, body values, assignee, and so on

test/support/test_helpers.ex
```diff
@@ -3,6 +3,7 @@ defmodule RemoteRetro.TestHelpers do
   alias RemoteRetro.{Repo, User, Vote}
```
vanderhoop: To avoid repeating `RemoteRetro`, you can leverage the `alias` macro, seen here.

test/support/test_helpers.ex
```diff
@@ -36,6 +37,27 @@ defmodule RemoteRetro.TestHelpers do
...
+    user_params = User.build_user_from_oauth(other_user)
+    other_user =
+      User.changeset(%User{}, user_params)
+      |> Repo.insert!
+
+    Map.put(context, :other_user, other_user)
+  end
+
+  def assign_idea(%{other_user: other_user, retro: retro} = context) do
+    idea = %RemoteRetro.Idea{assignee_id: other_user.id, body: "blurgh", category: "action-item", retro_id: retro.id, user_id: other_user.id} |> RemoteRetro.Repo.insert!
```
vanderhoop: Hey Nichol, tiny but helpful elixir tidbit is that you can leverage the `alias` macro (see comment) above so that you don't have to include the `RemoteRetro` module prefix repeatedly. To learn more, [peep this!](https://elixir-lang.org/getting-started/alias-require-and-import.html#alias)

test/support/test_helpers.ex
```diff
@@ -36,6 +37,27 @@ defmodule RemoteRetro.TestHelpers do
     Map.put(context, :user, user)
   end
 
+  def persist_other_user_for_retro(context) do
```
vanderhoop: hey @nicholalexander, i see what you mean about wanting to generalize user creation logic. given that each test helper takes and returns a context, maybe we could simply change the existing `persist_user_for_retro` to `persist_users_for_retro`, and have the callsites define `@users` as a list of users. That way, we can keep the logic in the existing (though altered) method and avoid introducing a new method just for this 'other user' case

test/support/test_helpers.ex
```diff
@@ -36,6 +37,27 @@ defmodule RemoteRetro.TestHelpers do
...
+
+    Map.put(context, :other_user, other_user)
+  end
+
+  def assign_idea(%{other_user: other_user, retro: retro} = context) do
+    idea = %RemoteRetro.Idea{assignee_id: other_user.id, body: "blurgh", category: "action-item", retro_id: retro.id, user_id: other_user.id} |> RemoteRetro.Repo.insert!
+    Map.put(context, :idea, idea)
+  end
+
+  def set_participation(%{other_user: other_user, retro: retro} = context) do
```
vanderhoop: given that a participation record represents the link of a user to a retro, i think we might want to encapsulate the creation of a participation in the `persist_users_for_retro` helper. the benefit is that then we don't have to call `set_participation` after we've called the helper that builds the necessary user. Whaddaya think?

browser_logs.log
```diff
@@ -0,0 +1,16 @@
+%cDownload the React DevTools for a better development experience: https://fb.me/react-devtools
```
vanderhoop: we recently added [`browser_logs.log` to the .gitignore](https://github.com/stride-nyc/remote_retro/blob/master/.gitignore#L45), so i think this branch may have been cut off of master before that change got in. @nicholalexander Can you rebase and repush to see if this file gets removed from the diff?

test/support/test_helpers.ex
nicholalexander: yes, i agree!  will update.

test/support/test_helpers.ex
nicholalexander: yessss.

test/support/test_helpers.ex
nicholalexander: 👍

test/support/test_helpers.ex
nicholalexander: cool.

test/support/test_helpers.ex
nicholalexander: the only reason not to do this was if we wanted to test users without participations - like potentially if a user was invited but never joined and was assigned an action item?  or something wierd like that?  Not sure.  Might be overly convoluted.  I think I'll just set it in `persist_users` as you suggest and then if its a problem down the road we can split it out.

browser_logs.log
nicholalexander: yes!  should'a seen this.

test/support/test_helpers.ex
vanderhoop: 👍 !

nicholalexander:
This refactors the `persist_idea` and `persist_users` to a general case but tries to keep previous use the same or as close to the same as possible.  I think overall, it's much clearer, except maybe the if else in `persist_idea_for_retro` - the else clause is the old way of doing which makes it so that the caller doesn't have to separately define an assignee.

I think this should be ok, and if the logic we use to assign ideas becomes more complex, we can revist how this works.

Primarily, you should check how things get called in the most complicated case, `when an action-item is reassigned` in `retro_idea_realtime_update_test.exs`.  

Additionally this adds Apex to the dependencies which is Elixir's version of Awesome Print.

config/test.exs
```diff
@@ -40,3 +40,14 @@ config :remote_retro, :mock_user, %{
   "profile" => "https://plus.google.com/108658712426577966861",
   "sub" => "108658712426577966861"
 }
+
+config :remote_retro, :other_user, %{
+  "email" => "misstestuser@gmail.com",
+  "email_verified" => "true", "family_name" => "Alexander",
+  "gender" => "female", "given_name" => "Nicole",
```
vanderhoop: Nicole, eh? 👍

test/support/test_helpers.ex
```diff
@@ -1,39 +1,65 @@
...
-      |> Repo.insert!
+    User.changeset(%User{}, user_params)
+        |> Repo.insert
+  end
+
+  def persist_users_for_retro(%{users: users} = context) do
+    Enum.each(users, fn user ->
+      persist_user(user)
+    end)
+    persisted_users = User |> Repo.all
```
vanderhoop: hey @nicholalexander! your `persist_user` helper returns a persisted user (with `id`) so here where you're doing `Enum.each`, you could instead do `Enum.map` and avoid the extra call to the db with `User |> Repo.all`

test/support/test_helpers.ex
```diff
@@ -1,39 +1,65 @@
...
   end
 
-  def persist_user_for_retro(context) do
-    context = Map.merge(%{user: @mock_user}, context)
-    %{user: user} = context
+  defp user_name_atom(name) do
+    String.replace(name, ~r/ +/, "") |> Macro.underscore |> String.to_atom
+  end
+
+  defp user_map(users) do
```
vanderhoop: I think we can do away with `user_name_atom` helper, as `Map`s can house string keys. This avoids the need for the helper, as well as the transformations done when building the map and accessing it. Separately, I think we should use the users `email` as the key in this map, rather than the user's name, as [email is unique](https://github.com/stride-nyc/remote_retro/blob/master/web/models/user.ex#L36).

test/support/test_helpers.ex
```diff
@@ -1,39 +1,65 @@
...
+    User.changeset(%User{}, user_params)
+        |> Repo.insert
+  end
+
+  def persist_users_for_retro(%{users: users} = context) do
+    Enum.each(users, fn user ->
+      persist_user(user)
+    end)
+    persisted_users = User |> Repo.all 
+    Map.merge(user_map(persisted_users), context)
```
vanderhoop: we'll want to reverse the arguments here, as we don't want any `users` on context taking precedent over the users you just persisted, as they'll have `id`s we want

test/support/test_helpers.ex
```diff
@@ -1,39 +1,65 @@
...
+  end
+
+  defp persist_assigned_idea(user, idea, retro) do
+    %Participation{retro_id: retro.id, user_id: user.id} |> Repo.insert!
```
vanderhoop: Hmmm... our application adds a `participation` record when the user joins a retro, so I think that the participation record should be persisted in the `persist_users_for_retro` helper, as it will confused anyone who doesn't have intimate knowledge of the system would be confused as to why a participation is being created as an implicit pre-req to the action item being assigned

test/features/retro_idea_realtime_update_test.exs
```diff
@@ -87,4 +92,33 @@ defmodule RetroIdeaRealtimeUpdateTest do
       assert String.contains?(action_items_list_text, "let's do the thing! (Test User)")
     end
   end
+
+  describe "when an action-item is reassigned" do
```
vanderhoop: test language here can be made a clearer: "when an action item is assigned"-> "it can be reassigned to another user"

nicholalexander: 👍

test/support/test_helpers.ex
nicholalexander: 💯

config/test.exs
nicholalexander: thought we'd want to have a good selection of test users.

test/support/test_helpers.ex
nicholalexander: I looked at this but there was a problem with the travis user sometimes already being in the database and couldn't quite figure out how that was happening.  i'm updating the persist to be an insert or find.

test/support/test_helpers.ex
nicholalexander: cool

test/support/test_helpers.ex
nicholalexander: i thought about that but i was afraid that would make the call sites in `retro_channel_test.exs` sort of wierd that you'd have to know the email address of your mock user to use it in the tests.  

also, i liked having the atom...

if anything i was thinking that we should refactor mock_user to be test_user so it's consistent with the name - or else change mock_user's name from test_user to mock_user (same difference really).  

I don't feel particularly strongly so up to you!

vanderhoop: > you'd have to know the email address of your mock user to use it in the tests

Good point. What if we changed the test users' names and emails to be something semantic, like `@test_user_one`, where the email is `testuser@one.com`, and `@test_user_two`, where the email is `testuser@two.com`. We could do the same thing with the names, as I think that would make things much clearer for all folks who ending up reading/writing tests?

nicholalexander: sounds good.  i'll make that change.

nicholalexander: made the change but left the names "Test User" and "Other User" because i felt it made it easier to read the error messages on failing tests.  I think it clarifies things, though.  This should be good to merge now.  Lemme know!

Issue #438, opened 2018-05-22
abort *.js commits that introduce js lint 


vanderhoop:
As a developer with staged *.js changes,
I want attempts to commit these files to be aborted if there's lint in the staged files,
So that lint doesn't sneak into the commit history.

__By Way Of Explanation:__ this story represents the first attempt to make our commit history cleaner and more consistent, alerting developers to lint that they've introduced.

__Acceptance Criteria:__

- [ ] attempts to commit staged *.js file changes are aborted if the changes contain lint

__Helpful Links:__

- [lint-staged](https://github.com/okonet/lint-staged)

vanderhoop:
implemented by @dplain90, closing!

PR #481, opened 2018-08-25
Workflow: Add precommit lint check #438

dplain90:
On commit attempt of *.js files, validates that changes against linter rules. If rules are not met, throws an error and cancels the commit - otherwise proceeds with the commit. 

- lint-staged - to kickoff eslint and validate
- husky - for git hooks

__Dependencies__ 
- lint-staged - ^7.2.2
- husky - ^0.14.3

[issue (#438)](https://github.com/stride-nyc/remote_retro/issues/438)


dplain90:
@vanderhoop Thanks!

> a) this needs to also run the linter on staged *.jsx files, as those make up the majority of our js codebase.

Cool got it.

> b) I think we should remove the git add portion, because I think lint could be introduced within a file that the developer only wants to commit a subset of the changes. if the lint isn't in the staged portion of the file, and it gets fixed and added, more might get added to the commit than necessary. does that make sense?

So you're talking about when someone is using ```git add -p``` on a file, similar to this [issue](https://github.com/okonet/lint-staged/issues/62) right?

Although there are some workarounds in the above issue, it doesn't look like there is any definitive one that hits all the edge cases. 

If we take out the git add, then the eslint autoformatting changes won't be staged after they are made. Alternatively, we can just avoid autoformatting/--fix all together for now, until the issue above is closed out. What do you think?

vanderhoop:
Hey @dplain90, that's exactly the issue I was worried about. looks like lots of other folks have run into the same issue. I _think_ we can keep the `--fix`, as the lint itself is what aborts the commit, and we'll want to fix that, even if the line isn't intended to go in the commit. And then the developer can then go in and `git add` anything that they definitely _do_ want in the commit.

dplain90:
@vanderhoop so your right that aborts it, but when ```--fix``` changes are run they won't be included in it's commit and will remain unstaged: 

1. Dev writes some code. 
2. Tries to commit. 
3. Eslint fixes styling errors and throws if it finds something wrong 
4. git add, re-adds files that eslint has fixed to the staging area so they are included( if git add command included)
5. Commit happens.  
 
So after step 3, eslint might have changes some parts of the file, but without the git add - they will just sit untracked which would mean you would have to do another commit just to add the autoformatting changes. You see what i mean?

vanderhoop:
Yeah, I follow. It's probably better to go with what you suggested earlier, killing both the `--fix` flag _and_ the `git add` call, just letting the dev handle the fixes and staging themselves. That way, nothing unwanted creeps into the diff, and the developer is (hopefully) trained to not introduce lint (or at least introduce less of it) in the first place, as they'll have to clean it up themselves. Whaddaya think, @dplain90?

dplain90:
@vanderhoop sounds good! I'll update.

package.json
```diff
@@ -2,15 +2,20 @@
   "repository": {},
   "license": "MIT",
   "scripts": {
+    "precommit": "lint-staged",
     "clean-hot-updates": "find . -name '*.hot-update*' -delete",
     "compile-production": "webpack -p --config='webpack.dll.js' --hide-modules && webpack -p",
     "compile-dev": "webpack --config='webpack.dll.js' --hide-modules && webpack --hide-modules",
     "watch": "webpack --config='webpack.dll.js' --hide-modules && NODE_ENV=dev node webpack.devserver.js",
+    "precommit": "lint-staged",
```
vanderhoop: hey @dplain90, overall the PR looks great. somehow this npm script addition was added twice, so gonna do a follow-up commit removing one them

PR #515, opened 2019-02-12
Favor `asdf` over `nvm` for version management of NodeJS

notactuallypagemcconnell:
`asdf` is an agnostic version management tool.
It would make sense to move to using it over `nvm` since that is already introduced for Elixir.

`asdf-nodejs` - added
`nvm` - removed

Ran on my system with exact copy/paste, installed the things, checked that it worked.

n/a

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,22 +76,17 @@ psql -h localhost
   - Create the "remote_retro_test" database and migrate via `MIX_ENV=test mix ecto.create && mix ecto.migrate`
 
 #### Node Dependencies
+We will use `asdf-nodejs`.
+It has a couple other dependencies, see [their README](https://github.com/asdf-vm/asdf-nodejs) for specifics for your system.
 
-Install nvm (node version manager):
-
+Install `nodejs` using `asdf` after following the directions from their README.
 ```
-curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
-```
-__Note:__ additional nvm installation notes found at [the nvm repo](https://github.com/creationix/nvm#install-script).
-
-Install the latest node via nvm:
-```
-nvm install 8.7
+asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs.git
 ```
 
-Ensure the latest node is your default node version in new shells:
+Install the project's version of nodejs
 ```
-nvm alias default 8.7
+asdf install nodejs $(cat .tool-versions | grep nodejs | awk '{print $2}')
 ```
 
 Install Global NPM Packages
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -77,21 +77,15 @@ psql -h localhost
...
-nvm install 8.7
+asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs.git
 ```
 
-Ensure the latest node is your default node version in new shells:
+Install the project's version
 ```
-nvm alias default 8.7
+asdf nodejs install 8.7.0
+asdf local nodejs 8.7.0 # use asdf global if you want it system-wide
```
notactuallypagemcconnell: I thought it could potentially be confusing for some users to have the system default set to this projects node version, so favored to keep it as a local configuration. I am totally open to changing this if anyone thinks there are downsides.

vanderhoop: Makes sense. Feels kinda dirty that this was our default from the beginning, and that's on me

README.md
```diff
@@ -77,21 +77,15 @@ psql -h localhost
...
 ```
-nvm install 8.7
+asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs.git
 ```
 
-Ensure the latest node is your default node version in new shells:
+Install the project's version
 ```
-nvm alias default 8.7
+asdf nodejs install 8.7.0
```
vanderhoop:
```suggestion
asdf install nodejs 8.7.0
```

README.md
```diff
@@ -77,21 +77,15 @@ psql -h localhost
 
 #### Node Dependencies
 
-Install nvm (node version manager):
-
-```
-curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.1/install.sh | bash
-```
-__Note:__ additional nvm installation notes found at [the nvm repo](https://github.com/creationix/nvm#install-script).
```
vanderhoop: https://github.com/asdf-vm/asdf-nodejs has a few other dependencies, so we might want to keep a note regarding more installation notes at their repo

README.md
vanderhoop: i think we can actually lock in the local `nodejs` version via the `.tool-versions` file, as asdf will read from that!

notactuallypagemcconnell: Oh good catch! Will add that.

README.md
notactuallypagemcconnell: Noted in the new amended commit but its not making this diff show outdated.

notactuallypagemcconnell:
@vanderhoop I think it should be all good now. One note that might slip through your eyes in the diff: I made the install line `asdf install nodejs $(cat .tool-versions | grep nodejs | awk '{print $2}')` because it will always grab the right one, but that might read a little strange to some. Thoughts?

Issue #519, opened 2019-02-19
upgrade phoenix 1.3 -> 1.4


vanderhoop:
As a developer,
I want to upgrade to the latest version of phoenix
So that we can remove code put in place to prevent race conditions that have been fixed in the latest framework.

__By Way Of Explanation:__ there was a race condition for folks leveraging the `Presence` api that's [been fixed in 1.4](https://gist.github.com/chrismccord/bb1f8b136f5a9e4abc0bfc07b832257e#update-your-presence-javascript)

__Acceptance Criteria:__

- [ ] all relevant upgrade steps in the guide below have been accounted for

__Helpful Links:__

- [Phoenix 1.3 -> 1.4 upgrade guide](https://gist.github.com/chrismccord/bb1f8b136f5a9e4abc0bfc07b832257e)

codyduval:
I'm working on this - I can't move any cards in ZenHub, but I'll update here when it's ready to go.

PR #524, opened 2019-03-13
Upgrade Phoenix to 1.4.1

codyduval:
Upgrade Phoenix and its related dependencies to version 1.4.1,
primarily to address race conditions within Phoenix Presence.

- Upgrade to Cowboy 2, which includes support for HTTP2
- Replace Poison with Jason for json generation
- Remove calls to transport macro, new api for socket config
- Refactor presence on client side to comply with new Presence API
- Update formatter config to format routing and test DSLs
- Alias Router.Helpers (instead of import) to remove compile-time dep
- Update circleci cache keys to force builds with new deps
- Fix bug in user_by_id reducer that was causing errors when
  users joined a retro.

Backend:
- `phoenix` 1.3.4 ---> 1.4.1
- `ecto_sql` 3.0 ---> 3.0.5
- `cowboy` 1.0 ---> `plug_cowboy` 2.0.1
- `plug` 1.7.2
- `jason` 1.1.2
- `poison` removed

Client:
- New `phoenix.js` included in phoenix package

__Relevant github Issue:__ 

- [Issue #519](https://github.com/stride-nyc/remote_retro/issues/519)


codyduval:
@vanderhoop I'll note that this should be heavily click-tested in some sort of staging environment before merging.

.env-sample
```diff
@@ -0,0 +1,3 @@
+export REMOTE_RETRO_GOOGLE_OAUTH_CLIENT_ID="<Client Id>"
```
vanderhoop: can we kill this? or include it in a separate PR with an explanation of its use? I'm not 100% against it; it just feels unrelated

.formatter.exs
```diff
@@ -1,9 +1,7 @@
...
-  inputs: [
-    "mix.exs",
-    "{config,lib,test}/**/*.{ex,exs}"
-  ],
+  import_deps: [:phoenix],
+  inputs: ["mix.exs", "*.{ex,exs}", "{config,lib,priv,test}/**/*.{ex,exs}"],
   line_length: 120,
   locals_without_parens: [],
-  trailing_comma: true,
+  trailing_comma: true
```
vanderhoop: the irony here is too good! love it!

.gitignore
```diff
@@ -47,3 +47,5 @@ package-lock.json
 npm-debug.log*
 .eslintcache
 browser_logs.log
+# Local secret env vars
```
vanderhoop: see comment above. also, if you want to make a case for inclusion in _this_ PR, i'm down to hear it!

.env-sample
vanderhoop: r, if you want to make a case for inclusion in this PR, i'm down to hear it!

web/static/js/services/retro_channel.js
```diff
@@ -11,27 +11,42 @@ class RetroChannel {
   }
 }
 
+export const handlePresence = (presence, setPresences) => {
```
vanderhoop: are we introducing this helper due to ease of testing? it almost feels like the indirection of passing the `presence` and `setPresences` action makes this a little less grokkable? 

what if we kept it inline down below?
```es6
const presence = new Presence(retroChannel)

presence.onSync(() => {
    const users = presence.list((_token, presence) => (presence.user))
    setPresences(users)
  })


```

test/redux/ideas_test.js
```diff
@@ -146,11 +146,12 @@ describe("actionCreators", () => {
 
         it("dispatches an error", () => {
           push = mockRetroChannel.push("anyEventJustNeedThePushInstance", { foo: "bar" })
+          push.trigger("ok", {})
```
vanderhoop: this is the only thing that gets me, as i can't understand *why* this is suddenly necessary. gonna peep through the phoenix source to see if there's something we can isolate in the `setupMockPhoenixChannel` helper

.env-sample
codyduval: Oh not a biggie, it can go elsewhere.

.gitignore
codyduval: 👍

web/static/js/services/retro_channel.js
codyduval: Yeah - originally it was more complicated, but now that it's a two liner, makes sense to get rid of it (I also thought I would test it separately, but that was also unneeded).

test/redux/ideas_test.js
codyduval: I know - I decided to ship it after too much time in the rabbit hole.  Would be great if you could provide insight.

vanderhoop: yeah, i've been rabbit holing for awhile, and it's baby steps

vanderhoop: by which i mean, slow and unsteady progress

vanderhoop:
great work on this, @codyduval! I'm going to merge this over the weekend, as it's a significant upgrade and changeset.

Issue #413, opened 2018-03-30
Add an Idea.changeset validation for the presence of an assignee_id when category is "action-item" 


vanderhoop:
At the moment, we enforce the selection of an assignee during the `"action-items"` stage, but this is only on the front-end. We never added the accompanying validation on our `Idea` struct in the backend, and we should to account for folks who get cute and somehow, some way, find a way to submit ideas through some backchannel.

__Helpful Links:__

- [`Ecto.Changeset` docs](https://hexdocs.pm/ecto/Ecto.Changeset.html)

samrrichards:
I just went through the setup process and would be happy to work on this issue when I get the chance!

vanderhoop:
hey @samrrichards, that would be a great help! thanks! let us know if you run into any issues! updated the ticket with the docs for Ecto changesets

samrrichards:
Great, will do!

codyduval:
I'll take this since it's been sitting for a bit.

PR #525, opened 2019-03-14
Validate presence of assignee_id on Idea when an action-item

codyduval:
When an `Idea` is created or modified, require that an `assignee_id` is present if the `Idea` is an `action-item`.

3 unit tests for new validation on `Idea` model.

- issue #413

### DIFF
--- a/test/models/idea_test.exs
+++ b/test/models/idea_test.exs
@@ -33,7 +33,7 @@ defmodule RemoteRetro.IdeaTest do
     end
 
     test "assignee_id NOT required when idea category NOT action-item" do
-      changeset = Idea.changeset(%Idea{category: "happy", assignee_id: 1})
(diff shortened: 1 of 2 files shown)


test/models/idea_test.exs
```diff
@@ -17,6 +17,28 @@ defmodule RemoteRetro.IdeaTest do
...
+    end
+
+    test "assignee_id and category action-item do not trigger error" do
+      changeset = Idea.changeset(%Idea{category: "action-item", assignee_id: 1})
+
+      refute Keyword.has_key?(changeset.errors, :assignee_id)
+    end
+
+    test "assignee_id NOT required when idea category NOT action-item" do
+      changeset = Idea.changeset(%Idea{category: "happy", assignee_id: 1})
```
vanderhoop: i think we'll want the `assignee_id` to be `nil` here! change that and we're golden!

Issue #453, opened 2018-06-27
make lobby guests wrap to multiple lines


williamcodes:
As a user
I want it to be more obvious that there are more users in the lobby than are visible
So that I don't delay starting because I think some people haven't joined yet

__By Way Of Explanation:__ 
This keeps confusing people at JOOR. We often have more participants than can fit on the screen. We often get first-time moderators and they never realize they can scroll. A simple fix would be allowing the list of faces to wrap to multiple lines when there are more profile picture thumbnails than will fit in a single line.

__Acceptance Criteria:__

- [x] the list of 'present' users wraps to multiple lines in the lobby and prime directive stage
   - see discussion below as to why we're scoping this wrapping to these stages

vanderhoop:
Hey @williamcodes,
Would like to groom this a bit, but curious about the scoping of this change to particular stages. For example, the following would be fine in the lobby/prime directive stages, but the real estate would be a bit much in the idea generation stage:

___

![image](https://user-images.githubusercontent.com/4386645/42631560-80cf6262-85a8-11e8-97ea-fed119b130a2.png)

My thinking is that we could make this change in the aforementioned stages, and down the road implement some sort of "expand" feature on the user list in the other stages, leaving the side scroll behavior there for now. Does that make sense?

williamcodes:
Yeah definitely, good thinking, I hadn't thought of the other stages of how
they would be impacted

vanderhoop:
done thanks to @edem-ca!

PR #528, opened 2019-03-18
Make user lobby guests wrap around

edem-ca:

Adds some CSS to allow lobby guests to wrap to multiple rows in the lobby and prime directive stages

`UserList` now takes a `wrap` property that toggles which form of the `UserList` you receive

image

https://github.com/stride-nyc/remote_retro/issues/453

### DIFF
--- a/web/static/js/components/css_modules/user_list.css
+++ b/web/static/js/components/css_modules/user_list.css
@@ -1,14 +1,6 @@
 :global(.basic.segment).index {
   :global(.horizontal.list) {
     margin-left: 0;
-    display: inline-grid;
-    overflow-x: scroll;
-    grid-template-rows: 1fr 1fr 1fr;
-    grid-auto-flow: column;
-
-    :global(.item), :global(.item:first-child){
-      margin: 0rem 1rem;
-    }
   }
   margin-top: 0;
   margin-bottom: 0;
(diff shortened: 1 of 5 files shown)


web/static/js/prop_types.js
```diff
@@ -58,3 +58,5 @@ export const votes = PropTypes.arrayOf(PropTypes.object)
 export const ideas = PropTypes.arrayOf(idea)
 
 export const actions = PropTypes.object
+
+export const wrap = PropTypes.bool
```
vanderhoop: we can omit adding `wrap` to this file , as it's more for prop types that are reused throughout the app, and `wrap` is only (currently) relevant to the `UserList` component

web/static/js/components/css_modules/user_list.css
```diff
@@ -12,3 +12,17 @@
   white-space: nowrap;
   overflow-x: scroll;
 }
+
+:global(.basic.segment).index {
+  :global(.wrap) {
+    display: inline-grid;
```
vanderhoop: hmmm... if we go this route, this will make the vertical columns fill up first. i think we can get away with a the rows wrapping wrapping by setting a max-height and overflow-y: scroll on the list element, and setting the list items to `display: inline-block`. That way, a new row will only be introduced once adding a new user caused a line break

Issue #489, opened 2018-09-06
allow users to access stage breakdowns after they've closed out the stage change alerts


vanderhoop:
As a user who would like to review the breakdown of a given stage,
I want a means of accessing said breakdown,
So that I can proceed confidently in the current sage.

__By Way Of Explanation:__ users often close out the 'Stage Change' alerts containing stage breakdowns/guidance, only to then want to be able to review the rules of the current stage, but there is currently no way of doing so. yet 😄 

__Acceptance Criteria:__

- [ ] there is a SemanticUI "question circle" icon in the upper right hand of the header, to the left of the 'Built by Stride' logo
- [ ] upon hovering over the "question circle" icon for a moment, the native HTML title reads, "Access Stage Information"
- [ ] upon clicking the "question circle" icon, the same alert modal content that's surfaced during stage changes is surfaced.
  - __Note:__ do not include the 'Stage Change: [CURRENT_STAGE]' header seen in the stage change alerts, as the user will already be there
- [x] __A Nice to Have:__ upon exiting out of the Stage Change alert, this icon fades in and out three times to draw the user's eye to it

__Helpful Links:__

- [Semantic UI Icons](https://semantic-ui.com/elements/icon.html)
- [React Portals](https://reactjs.org/docs/portals.html)
  - the header element exists outside the scope of our normal React component heirarchy. React 16.x exposes "portals" for dealing with elements outside of said hierarchy.
- [The 'alert' reducer case that surfaces stage change alerts](https://github.com/stride-nyc/remote_retro/blob/master/web/static/js/redux/alert.js#L11-L14) and the [StageConfigs](https://github.com/stride-nyc/remote_retro/blob/e410fe7c80e59fd0ad1bf4301469958f6df51ea9/web/static/js/configs/stage_configs.jsx#L21-L24) map that it consumes

codyduval:
I've got some time in the mornings this week - I'll see if I can make some progress on this one.

PR #530, opened 2019-03-21
Allow users to access stage breakdowns for help

codyduval:

- Adds a question mark icon to the header which, when clicked,
invokes the stage breakdown modal for the current retro stage.

- Adds a new redux action for alerts to invoke help.

- Adds a new 'help' key to the stageConfig.js to show abbreviated
header for the help modal (does not show "Stage Change! when it
is help).

- Since the header exists outside of the react dom, uses a React
portal to render the question icon.

- Quesion mark icon renders conditionally based on stage, only
showing when there is help to be shown.

Enzyme tests for new `StageHelp` component.

- Issue #489

### DIFF
--- a/test/components/stage_help_test.js
+++ b/test/components/stage_help_test.js
@@ -3,6 +3,7 @@ import sinon from "sinon"
 import { shallow } from "enzyme"
 
 import { StageHelp } from "../../web/static/js/components/stage_help"
+import STAGES from "../../web/static/js/configs/stages"
 
 describe("<StageHelp />", () => {
   let wrapper
@@ -25,7 +26,7 @@ describe("<StageHelp />", () => {
 
   describe("when it is a stage with help to show", () => {
     beforeEach(() => {
-      retro = { stage: "idea-generation" }
+      retro = { stage: STAGES.IDEA_GENERATION }
       const newProps = { ...defaultProps, retro }
       wrapper = shallow(<StageHelp {...newProps} />)
     })
@@ -44,7 +45,7 @@ describe("<StageHelp />", () => {
 
   describe("when it is a stage with no help to show", () => {
     beforeEach(() => {
-      const retro = { stage: null }
+      const retro = { stage: STAGES.LOBBY }
       const newProps = { ...defaultProps, retro }
       wrapper = shallow(<StageHelp {...newProps} />)
     })
--- a/test/components/stage_help_test.js
+++ b/test/components/stage_help_test.js
@@ -1,0 +1,56 @@
+import React from "react"
+import sinon from "sinon"
+import { shallow } from "enzyme"
+
+import { StageHelp } from "../../web/static/js/components/stage_help"
+
+describe("<StageHelp />", () => {
+  let wrapper
+  let retro
+
+  // Setup JSDom so that react can inject a portal
+  const iconRoot = global.document.createElement("div")
+  iconRoot.setAttribute("id", "stage-help-icon")
+  const body = global.document.querySelector("body")
+  body.appendChild(iconRoot)
+
+  const actions = {
+    showStageHelp: sinon.spy(),
+  }
+
+  const defaultProps = {
+    actions,
+    retro: {},
+  }
+
+  describe("when it is a stage with help to show", () => {
+    beforeEach(() => {
+      retro = { stage: "idea-generation" }
+      const newProps = { ...defaultProps, retro }
+      wrapper = shallow(<StageHelp {...newProps} />)
+    })
+
+    it("renders the question mark icon", () => {
+      expect(wrapper.find("i.question").exists()).to.equal(true)
+    })
+
+    describe("when clicking the icon for help", () => {
+      it("dispatches showStageHelp with the retro object", () => {
+        wrapper.find("i.question").simulate("click")
+        expect(actions.showStageHelp).to.have.been.calledWith(retro)
+      })
+    })
+  })
+
+  describe("when it is a stage with no help to show", () => {
+    beforeEach(() => {
+      const retro = { stage: null }
+      const newProps = { ...defaultProps, retro }
+      wrapper = shallow(<StageHelp {...newProps} />)
+    })
+
+    it("does NOT render the question mark icon", () => {
+      expect(wrapper.find("i.question").exists()).to.equal(false)
+    })
+  })
+})
--- a/test/redux/alert_test.js
+++ b/test/redux/alert_test.js
@@ -59,6 +59,39 @@ describe("alert", () => {
       })
     })
 
+    describe("when the action is SHOW_STAGE_HELP", () => {
+      const initialState = { headerText: "Warning!", bodyText: "You're being watched." }
+      const stageConfigs = {
+        daybreak: {
+          help: {
+            headerText: "I need help!",
+            bodyText: "Tell me what to do",
+          },
+        },
+        stageLackingHelpConfig: {},
+      }
+
+      deepFreeze(initialState)
+      deepFreeze(stageConfigs)
+
+      describe("when the given stage has a help in the given configuration map", () => {
+        const action = {
+          type: "SHOW_STAGE_HELP",
+          retro: {
+            stage: "daybreak",
+          },
+          stageConfigs,
+        }
+
+        it("returns the help for the given stage", () => {
+          expect(reducer(initialState, action)).to.deep.equal({
+            headerText: "I need help!",
+            bodyText: "Tell me what to do",
+          })
+        })
+      })
+    })
+
     describe("when the action is CLEAR_ALERT", () => {
       const action = { type: "CLEAR_ALERT" }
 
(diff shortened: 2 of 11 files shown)


web/static/js/configs/stage_configs.jsx
```diff
@@ -21,6 +21,10 @@ const baseIdeaGenerationConfig = {
     headerText: "Stage Change: Idea Generation!",
     BodyComponent: StageChangeInfoIdeaGeneration,
   },
+  help: {
+    headerText: "Idea Generation",
+    BodyComponent: StageChangeInfoIdeaGeneration,
+  },
```
codyduval: @vanderhoop There were a few ways to solve the "Don't show the Stage Change!" text requirement, I chose to introduce some duplication in the config object instead of some sort of conditional in the component.  Pros and cons with each approach, but I felt was the cleanest.

test/redux/alert_test.js
```diff
@@ -59,6 +59,39 @@ describe("alert", () => {
...
+            bodyText: "Tell me what to do",
+          },
+        },
+        stageLackingHelpConfig: {},
+      }
+
+      deepFreeze(initialState)
+      deepFreeze(stageConfigs)
+
+      describe("when the given stage has a help in the given configuration map", () => {
```
vanderhoop: a thought: "do we need a test to handle the other case?"

web/static/js/components/stage_help.jsx
```diff
@@ -0,0 +1,62 @@
...
+        />
+      </div>,
+      document.getElementById("stage-help-icon")
+    )
+  }
+
+  return null
+}
+
+const mapStateToProps = state => {
```
vanderhoop: I don't think this component will need to be `connect`ed, as it's parent, the `<RemoteRetro />` component, already has the `retro` state and it can pass it down, right?

web/static/js/redux/alert.js
```diff
@@ -1,5 +1,7 @@
 const types = {
   CLEAR_ALERT: "CLEAR_ALERT",
+  RETRO_UPDATE_COMMITTED: "RETRO_UPDATE_COMMITTED",
```
vanderhoop: I realize this is a refactor, but if we're going to clean this up, would it be better to import the `types` from the "retro" redux file?

web/static/js/prop_types.js
```diff
@@ -58,3 +58,11 @@ export const votes = PropTypes.arrayOf(PropTypes.object)
 export const ideas = PropTypes.arrayOf(idea)
 
 export const actions = PropTypes.object
+
```
vanderhoop: ❤️ thank you for doing this

web/static/js/configs/stage_configs.jsx
vanderhoop: yeah, i hear that. think this is a sensible choice.

web/static/js/components/stage_help.jsx
```diff
@@ -0,0 +1,62 @@
...
+export const StageHelp = props => {
+  const handleClick = () => {
+    const { retro, actions: { showStageHelp } } = props
+    showStageHelp(retro)
+  }
+
+  const showIcon = () => {
+    const { retro: { stage } } = props
+    const validHelpStages = [
+      STAGES.PRIME_DIRECTIVE,
```
vanderhoop: hmmm... this datastructure is implicitly coupled to the stage config object, as the only stage omitted here is the one state where `help` is null. so maybe we could just key into that object and `showIcon` if and only if `!help`?

test/components/stage_help_test.js
```diff
@@ -0,0 +1,56 @@
...
+
+describe("<StageHelp />", () => {
+  let wrapper
+  let retro
+
+  // Setup JSDom so that react can inject a portal
+  const iconRoot = global.document.createElement("div")
+  iconRoot.setAttribute("id", "stage-help-icon")
+  const body = global.document.querySelector("body")
+  body.appendChild(iconRoot)
```
vanderhoop: dude, nice work on these tests. have to imagine it took some digging to get the setup right.

web/static/js/redux/alert.js
vanderhoop: and in fact, `SHOW_STAGE_HELP` is also in the `retros` redux file's `types`, so if we *can* import them, I think we should house both there. any reason we wouldn't? not sure if there's a redux convention for this

web/static/js/components/stage_help.jsx
codyduval: Yes, it looks like the `retro` state is in the parent - good catch.  But I still need to `dispatch` from the component so I'll need to `connect`.

test/redux/alert_test.js
codyduval: Well, the other case will return an `undefined`. I can make the reducer return `null` instead and then the `Alert` component should just not render if it hits that condition.

web/static/js/redux/alert.js
codyduval: Hmm - convention is all over the place here.  Since we're housing reducers and actionCreators together, we should probably split the types.  So SHOW_STAGE_HELP and CLEAR_ALERT stay in the alert redux file, and the others stay in the retros file.  We then import in both files...

web/static/js/components/stage_help.jsx
codyduval: Cool - good idea.

codyduval: The internet knows all.

test/components/stage_help_test.js
codyduval: The internet knows all.

test/redux/alert_test.js
vanderhoop: i believe the `Alert` component already does a null render when there's no `alert` object passed as a prop, so we're good there

web/static/js/components/stage_help.jsx
```diff
@@ -0,0 +1,48 @@
...
+        />
+      </div>,
+      document.getElementById("stage-help-icon")
+    )
+  }
+
+  return null
+}
+
+const mapDispatchToProps = dispatch => ({
```
vanderhoop: @codyduval the `<RemoteRetro />` component also has bound `actions`, so i think you can kill the `connect` call outright and pass the `actions` object down

vanderhoop: peep https://github.com/stride-nyc/remote_retro/blob/master/web/static/js/components/remote_retro.jsx#L100-L102

web/static/js/components/stage_help.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const StageHelp = props => {
+  const handleClick = () => {
+    const { retro, actions: { showStageHelp } } = props
+    showStageHelp(retro)
+  }
+
+  const showIcon = () => {
+    const { retro: { stage } } = props
+    return (stageConfig[stage] && stageConfig[stage].help)
```
vanderhoop: we should be able to axe the short-circuiting here, as each `stage` is a constant, and are all accounted for in the `stageConfig` object. While in theory they _can_ change, any unaccounted for changes would result in a number of end-to-end tests failing, so we shouldn't need to be defensive here. unless i'm missing something?

web/static/js/redux/alert.js
```diff
@@ -1,5 +1,8 @@
-const types = {
+import { types as retroTypes } from "./retro"
+
+export const types = {
   CLEAR_ALERT: "CLEAR_ALERT",
+  SHOW_STAGE_HELP: "SHOW_STAGE_HELP",
```
vanderhoop: ❤️

web/static/js/components/stage_help.jsx
codyduval: Hmm - I've never passed an action creator already bound to dispatch.  Lemme see if that works.

web/static/js/components/stage_help.jsx
codyduval: Sure - it's just such a common error I just do it reflexively.

test/redux/alert_test.js
```diff
@@ -59,6 +59,56 @@ describe("alert", () => {
...
+      describe("when the given stage does NOT have a help in the given configuration map", () => {
+        const action = {
+          type: "SHOW_STAGE_HELP",
+          retro: {
+            stage: "nightfall",
+          },
+          stageConfigs,
+        }
+
+        it("returns the help for the given stage", () => {
```
vanderhoop: we'll need to update the test language here to reflect that we expect null

web/static/js/components/stage_help.jsx
codyduval: You're right, the actions can get passed down.

web/static/js/components/stage_help.jsx
codyduval: Yup works.

test/redux/alert_test.js
codyduval: 👍

vanderhoop:
dude, really nice work on this. mergin!

PR #533, opened 2019-04-04
Transfer facilitatorship

edem-ca:

If I am the facilitator of a retro, I want to be able to make someone else the facilitator!

Nothing new on the elixir side, some new functionality was added to the `alert` and `retro` reducers were added so there are some new tests there.

image
image
image
image

image

image

image

https://app.zenhub.com/workspaces/remote-retro-5874435a4a416214111ed3be/issues/stride-nyc/remote_retro/190


test/components/remote_retro_test.js
```diff
@@ -32,4 +39,28 @@ describe("RemoteRetro component", () => {
...
+
+    it("calls the new facilitator action", () => {
+      const wrapper = shallow(<RemoteRetro {...defaultProps} />)
+      wrapper.setProps({ user: newFacilitator })
+      expect(defaultProps.actions.newFacilitator.calledWith(defaultProps.retro))
+    })
+  })
+
+  describe("when component updates new stage", () => {
+    it("does not the new facilitator action", () => {
```
vanderhoop: based on the implementation code below, i think the test setup here is a little misleading, as the condition checks whether there is specifically a change in the currentUser's `is_facilitator` boolean attribute (from *not* being facilitator *to* being facilitator). So in the `setProps()` call, we would make the component update by including a currentUser who is still *not* the facilitator

test/components/remote_retro_test.js
```diff
@@ -32,4 +39,28 @@ describe("RemoteRetro component", () => {
...
+    })
+
+    it("calls the new facilitator action", () => {
+      const wrapper = shallow(<RemoteRetro {...defaultProps} />)
+      wrapper.setProps({ user: newFacilitator })
+      expect(defaultProps.actions.newFacilitator.calledWith(defaultProps.retro))
+    })
+  })
+
+  describe("when component updates new stage", () => {
```
vanderhoop: per my comment below, I think we'll want to update the `describe` language here to reflect that the conditions we're setting up is a non-change to the current user's facilitatorship

test/components/remote_retro_test.js
```diff
@@ -32,4 +39,28 @@ describe("RemoteRetro component", () => {
...
+    it("does not the new facilitator action", () => {
+      const wrapper = shallow(<RemoteRetro {...defaultProps} />)
+      wrapper.setProps({ stage: CLOSED })
+      expect(defaultProps.actions.newFacilitator.notCalled)
```
vanderhoop: will this test always pass? I think we need an assertion attached to the `expect()`, but i could be wrong.

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,57 @@ describe("UserListItem", () => {
     })
   })
 
+  describe("passed a non-facilitator as a facilitator", () => {
```
vanderhoop: "passed a non-facilitator as a facilitator" -> "passed a non-facilitator user"

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,57 @@ describe("UserListItem", () => {
...
+    const user = { ...defaultUserAttrs, is_facilitator: true }
```
vanderhoop: a nitpick, but you `user` -> `facilitator` and `user2` -> `nonFacilitator` would make this easier to read, as then readers don't need to remember that user one is the facilitator and user two isn't

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,57 @@ describe("UserListItem", () => {
...
+    const user2 = { ...secondaryrUserAttrs, is_facilitator: false }
+
+    it("renders the magic icon", () => {
+      const wrapper = shallow(
+        <UserListItem {...defaultProps} user={user2} currentUser={user} />
+      )
+      expect(wrapper.find("i.magic.icon")).to.have.length(1)
+    })
+
+    it("renders facilitator button", () => {
```
vanderhoop: __Note to self:__ check if these are separate buttons in implementation

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,57 @@ describe("UserListItem", () => {
...
+    const user = { ...defaultUserAttrs, is_facilitator: true }
+
+    it("does not render the magic icon", () => {
+      const wrapper = shallow(
+        <UserListItem {...defaultProps} user={user} currentUser={user} />
+      )
+      expect(wrapper.find("i.magic.icon")).to.have.length(0)
+    })
+
+    it("does not render the magic icon", () => {
```
vanderhoop: the test language here is a dupe of the test above, but we're asserting that there is a facilitator button

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,57 @@ describe("UserListItem", () => {
...
+    it("renders facilitator button", () => {
+      const wrapper = shallow(
+        <UserListItem {...defaultProps} user={user2} currentUser={user} />
+      )
+      expect(wrapper.find("button.facilitator")).to.have.length(1)
+    })
+
+    describe("passing facilitatorship", () => {
+      beforeEach(() => {
+        window.confirm = sinon.stub().returns(true)
```
vanderhoop: we'll need to add a test to cover the case that the user *doesn't* confirm.

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,57 @@ describe("UserListItem", () => {
...
+      })
+
+      it("calls the action to change facilitator", () => {
+        const wrapper = shallow(
+          <UserListItem {...defaultProps} user={user2} currentUser={user} />
+        )
+        wrapper.find("button.facilitator").simulate("click")
+        expect(defaultProps.actions.updateRetroAsync.calledWith({ facilitator_id: user2.id }))
```
vanderhoop: same as above. curious if this will always fail, as there is no assertion tagged on the `expect()`?

web/static/js/components/user_list_item.jsx
```diff
@@ -8,18 +10,35 @@ import STAGES from "../configs/stages"
...
   return (
     <li className={`item ${styles.wrapper}`}>
+      {renderButton
+      && (
+        <button
+          type="button"
+          className={`${styles.facilitator}`}
+          onClick={() => passFacilitatorship(user, currentUser)}
+        >
+          {currentUser !== user && <i className="magic icon" />}
```
vanderhoop: i think when we extract this, we'll want to do a comparison of the `currentUser.id` and `user.id`, as that will protect us if they are two separate objects that are identical in contents, which is possible, especially in redux apps due to the preferring transformations to mutations

web/static/js/redux/alert.js
```diff
@@ -12,13 +14,20 @@ export const actions = {
...
-      return stageConfigs[retro.stage].alert || null
+      const { retroChanges, stageConfigs } = action
+      return retroChanges.stage ? stageConfigs[retroChanges.stage].alert : null
     }
     case types.SHOW_STAGE_HELP: {
       const { retro, stageConfigs } = action
       return stageConfigs[retro.stage].help || null
     }
+    case types.NEW_FACILITATOR: {
+      const configs = {
```
vanderhoop: (nitpick) we can return the object without putting it in a `const`

web/static/js/components/user_list_item.jsx
```diff
@@ -8,18 +10,35 @@ import STAGES from "../configs/stages"
 
 const { VOTING } = STAGES
 
-export const UserListItem = ({ user, votes, stage }) => {
+export const UserListItem = ({ user, votes, stage, currentUser, actions }) => {
   const imgSrc = user.picture.replace("sz=50", "sz=200")
   const votesByUser = votes.filter(vote => vote.user_id === user.id).length
   const allVotesIn = votesByUser >= VOTE_LIMIT
+  const renderButton = !user.is_facilitator && currentUser.is_facilitator
```
vanderhoop:
```suggestion
renderButton -> renderFacilitatorTransferButton 
```

or something that identifies the button's purpose in an expressive way

web/static/js/components/user_list_item.jsx
edem-ca: hadn't considered that 🤔 good looks

test/components/remote_retro_test.js
edem-ca: we do, an oversight on my part

test/components/user_list_item_test.js
edem-ca: the button contains the icon inside of it

That logic is there so that the magic icon doesn't render for your _own_ avatar. Not that it's particularly harmful to pass facilitstorship to yourself, but in the moment it made more sense to put a bit of logic to prevent it

test/components/remote_retro_test.js
edem-ca: agreed

test/components/user_list_item_test.js
```diff
@@ -45,6 +59,81 @@ describe("UserListItem", () => {
...
+          { facilitator_id: nonFacilitator.id }
+        )).to.eql(true)
+      })
+
+      afterEach(() => {
+        defaultProps.actions.updateRetroAsync.reset()
+      })
+    })
+
+    describe("declining the passing of facilitatorship", () => {
```
vanderhoop: ❤️ 😍 💓

web/static/js/components/user_list_item.jsx
```diff
@@ -8,18 +10,38 @@ import STAGES from "../configs/stages"
...
+      && (
+        <button
+          type="button"
+          data-tooltip="Transfer facilitatorship"
+          data-inverted=""
+          data-position="bottom right"
+          className={`${styles.facilitator}`}
+          onClick={() => passFacilitatorship(user, currentUser)}
+        >
+          {currentUser.id !== user.id && <i className="ui magic icon" />}
```
vanderhoop: i still think this might be unnecessary, as the condition above checks whether the current user *is* the facilitator and we only show the button if the user being rendered in the list item *isn't* the facilitator, so their ids would never match. Right? I'll check.

web/static/js/components/remote_retro.jsx
```diff
@@ -23,8 +24,11 @@ export class RemoteRetro extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { stage } = this.props
+    const { stage, currentUser, actions, retro } = this.props
     if (prevProps.stage !== stage) { hj("trigger", stage) }
+    if (!prevProps.currentUser.is_facilitator && currentUser.is_facilitator) {
+      actions.newFacilitator(retro)
```
vanderhoop: no need to pass `retro` here, based on the method signature. will remove

web/static/js/components/user_list_item.jsx
vanderhoop: double checked, and a couple things came up here. 1) is that we can safely remove the short circuit in `currentUser.id !== user.id` and still get the desired behavior due to the checks described in the above comment, and 2) is that removing the check didn't cause any tests to fail, so make sure you're always test driving to make sure you're testing what you think they are.

test/components/remote_retro_test.js
```diff
@@ -32,4 +39,28 @@ describe("RemoteRetro component", () => {
...
+      newFacilitator = {
+        ...stubUser,
+        is_facilitator: true,
+      }
+    })
+
+    it("calls the new facilitator action", () => {
+      const wrapper = shallow(<RemoteRetro {...defaultProps} />)
+      wrapper.setProps({ user: newFacilitator })
+      expect(defaultProps.actions.newFacilitator.calledWith(defaultProps.retro))
```
vanderhoop: @edem-ca hey bud! just a heads up that this was always passing, do to there being no assertion attached to the expect. I removed the passage of `retro` in the implementation, and it was still green. Again, this surfaces a need for you to test drive your changes, as that would have made the false positive apparent!

vanderhoop: also! there was an extension of the false positive issue, in that the `setProps` call wasn't actually triggering the code in question, due to passing `user` instead of `currentUser`

test/components/remote_retro_test.js
```diff
@@ -8,17 +10,22 @@ const { IDEA_GENERATION, CLOSED } = STAGES
...
     retroChannel: mockRetroChannel,
     isTabletOrAbove: true,
     presences: [],
-    actions: {},
     ideas: [],
     stage: IDEA_GENERATION,
     facilitatorName: "Daniel Handpan",
     retro: { stage: IDEA_GENERATION },
+    actions: {
+      newFacilitator: spy(),
```
vanderhoop: gonna remove the spy from default props, as this gets applied to all of the tests, and shift spy declaration down into a `beforeEach` so that the spies reset

PR #550, opened 2019-11-08
Change "you're" to "we're" to sound more inclusive and welcoming.

piannone-orchard:
Change "you're" to "we're" to sound more inclusive and welcoming.


vanderhoop:
❤️!

Issue #435, opened 2018-05-16
notify facilitator if all votes are cast or not as an aggregate


williamcodes:
As a user,
I want an easy way to see if it's time to end the voting phase
So that I don't have to check all the vote cast statuses for participants, particularly in large groups where not all are visible at once.

__By Way Of Explanation:__ (optional)
Roberto Pedroso (lead eng for JOOR) requested this while running a retro

__Acceptance Criteria:__

- [ ] in the voting stage, when every user currently present in a retro has submitted all of their votes, a teal, Semantic UI 'pointing label' appears above the stage progression button, reading "All votes in! Let's move forward!"

__Should look something like this:__ 
![image](https://user-images.githubusercontent.com/4386645/42631832-58f755d2-85a9-11e8-900f-120f6b6e79e3.png)

__Helpful Links:__

- [Semantic UI Pointing Label](https://semantic-ui.com/elements/label.html)

vanderhoop:
I think this is a great feature idea. We could prompt the facilitator to click "Proceed to Action Items" with a pointing tooltip that reads something like, "All votes in!", or remove the middleman and just progress the retro automatically once all participants have used up their votes. That might be a bit much; I don't know.

vanderhoop:
example tooltip prompt: 

![image](https://user-images.githubusercontent.com/4386645/42631832-58f755d2-85a9-11e8-900f-120f6b6e79e3.png)

vanderhoop:
hey @williamcodes. groomed this one! I'll throw it in the bench channel

vanderhoop:
closing, as implemented by @pwjablonski in 8dfac0d

PR #551, opened 2019-11-15
notifies facilitator if all votes are cast - Addresses issue #435

pwjablonski:
Hey! 

I'm not sure if this issue has already been claimed but thought I'd give it a shot! I'm looking to contribute to more open source projects and this looked liked a great place to start. I had a bunch of fun navigating the code and implementing this feature. I'm very open to your feedback and suggestions on the approach here. I have a couple of line-level questions that I've added in comments.
Peter

 - [x] in the voting stage, when every user currently present in a retro has submitted all of their votes, a teal, Semantic UI 'pointing label' appears above the stage progression button, reading "All votes in!"

- N/A

- No tests added yet

Screen Shot 2019-11-14 at 6 45 48 PM

- [#435](https://github.com/stride-nyc/remote_retro/issues/435)

### DIFF
--- a/test/components/stage_progression_button_test.js
+++ b/test/components/stage_progression_button_test.js
@@ -153,6 +153,53 @@ describe("StageProgressionButton", () => {
         expect(stageProgressionButton.find("div")).to.have.length(0)
       })
     })
+
+    context("when stateDependentTooltip returns a value", () => {
+      beforeEach(() => {
+        const props = {
+          ...defaultProps,
+          config: { ...config, stateDependentTooltip: () => "All in!" },
+        }
+        stageProgressionButton = mountWithConnectedSubcomponents(
+          <StageProgressionButton {...props} />
+        )
+      })
+
+      it("renders tooltip", () => {
+        expect(stageProgressionButton.find(".tooltip").text()).to.equal("All in!")
+      })
+    })
+
+    context("when stateDependentTooltip does not return a value", () => {
+      beforeEach(() => {
+        const props = {
+          ...defaultProps,
+          config: { ...config, stateDependentTooltip: () => null },
+        }
+        stageProgressionButton = mountWithConnectedSubcomponents(
+          <StageProgressionButton {...props} />
+        )
+      })
+
+      it("does not render tooltip", () => {
+        expect(stageProgressionButton.find(".tooltip")).to.have.length(0)
+      })
+    })
+
+    context("when stateDependentTooltip does not exisit", () => {
+      beforeEach(() => {
+        const props = {
+          ...defaultProps,
+        }
+        stageProgressionButton = mountWithConnectedSubcomponents(
+          <StageProgressionButton {...props} />
+        )
+      })
+
+      it("does not render tooltip", () => {
+        expect(stageProgressionButton.find(".tooltip")).to.have.length(0)
+      })
+    })
   })
 
   describe("when it does not receive a progressionButton configuration object", () => {
--- a/web/static/js/components/stage_progression_button.jsx
+++ b/web/static/js/components/stage_progression_button.jsx
@@ -1,15 +1,12 @@
 import React, { Component } from "react"
 import { connect } from "react-redux"
 import { bindActionCreators } from "redux"
-import isEmpty from "lodash/isEmpty"
 
 import PropTypes from "prop-types"
 import Modal from "react-modal"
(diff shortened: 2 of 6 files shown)


web/static/js/components/stage_progression_button.jsx
```diff
@@ -33,11 +36,15 @@ export class StageProgressionButton extends Component {
...
+      presences,
       retroUpdateRequested,
       config,
+      votes,
     } = this.props
 
-    const { modalOpen, showSubmitIdeaPrompt } = this.state
+    const { modalOpen } = this.state
+    const showAllVotesInPrompt = !isEmpty(presences)
+      && (votes.length >= VOTE_LIMIT * presences.length)
```
pwjablonski: We calculate the total number of maximum possible votes across all users ( `presences.length` * `VOTE_LIMIT`) as well as the current number of votes cast so far (`votes.length`). If they are equal we know to display the tooltip.

web/static/js/components/stage_progression_button.jsx
```diff
@@ -33,11 +36,15 @@ export class StageProgressionButton extends Component {
...
       currentUser,
+      presences,
       retroUpdateRequested,
       config,
+      votes,
     } = this.props
 
-    const { modalOpen, showSubmitIdeaPrompt } = this.state
+    const { modalOpen } = this.state
+    const showAllVotesInPrompt = !isEmpty(presences)
```
pwjablonski: We are determining we are on the correct the stage (voting stage) here by the lack of presences. Would this be better determined in another way?

web/static/js/components/stage_progression_button.jsx
```diff
@@ -111,6 +121,7 @@ StageProgressionButton.defaultProps = {
 const mapStateToProps = reduxState => ({
   reduxState,
   retroUpdateRequested: reduxState.retro.updateRequested,
+  votes: reduxState.votes,
```
pwjablonski: Followed the prior art in the line above, but we could also pull this out into a selector if preferable?

vanderhoop: we can leave these as as, as there's not much computation/logic to these lines.

web/static/js/components/stage_progression_button.jsx
vanderhoop: I think this introduces a bug, in that the tooltip would show in all stages _following_ the voting stage.

Thankfully, we can scope this to the voting stage, as we have the `stage` available in the redux store (under the `retro` key/reducer), so if we pass the stage in the `connect` call, that simplifies this logic and scopes the tooltip to this stage alone.

pwjablonski:
Hey @vanderhoop! Thanks for the feedback - that makes sense to me! I went ahead and made changes based on your suggestions. Let me know if this approach is more consistent with what you are thinking. I also added some tests, though I would love some guidance on crafting them. Looking forward to your review!

web/static/js/configs/happy_sad_confused_stage_configs.js
```diff
@@ -112,6 +113,13 @@ export default {
       copy: "Proceed to Action Items",
       iconClass: "arrow right",
       confirmationMessage: "Are you sure you would like to proceed to the action items stage?",
+      stateDependentTooltip: reduxState => {
+        const { votes, presences } = reduxState
+        if (votes.length >= VOTE_LIMIT * presences.length) {
```
vanderhoop: @pwjablonski This isn't a blocker to merge, but we can make this inline logic more immediately grokkable to future maintainers by extracting `VOTE_LIMIT * presences.length` out into a constant that expresses the concept. Something like, `TOTAL_VOTE_LIMIT_FOR_RETRO`.

Again, not a blocker. I'll add this after the fact.

test/components/stage_progression_button_test.js
```diff
@@ -153,6 +153,53 @@ describe("StageProgressionButton", () => {
         expect(stageProgressionButton.find("div")).to.have.length(0)
       })
     })
+
+    context("when stateDependentTooltip returns a value", () => {
```
vanderhoop: @pwjablonski These tests are on point. Thank you for being thorough with the various cases handled.

pwjablonski:
@vanderhoop awesome - thanks for the feedback! Fun to see the changes live!

Issue #556, opened 2020-01-30
provide UI feedback to facilitators when group.label updates have been successfully persisted


vanderhoop:
As a facilitator who has submitted a group.label update,
I want to know whether the update is successful,
So that I can have faith that the update was persisted and I can move on to updating the next group.label.

__By Way Of Explanation:__ at the moment, when a facilitator submits an update to a group.label, we broadcast the update to all connected clients, and the redux store is updated appropriately. However, the facilitator has no way of knowing that their update went through. This story is about adding a small amount of UI feedback to this experience so that there is no question/nag in the user's mind after an update.

__Acceptance Criteria:__

- [ ] when a group.name is successfully updated in the redux store, a green, SemanticUI checkmark is surfaced __to the right of the facilitator's input__
  - [ ] said checkmark is removed after 1s 
  - __Note:__ non-facilitator's have a readonly paragraph tag for the value, and this is updated when the redux store's value is updated. This story is just about providing feedback to the facilitator beside their input interface.

__Helpful Links:__

- [uncontrolled and controlled form inputs in React](https://reactjs.org/docs/uncontrolled-components.html)
  - __Note:__ the article's title implies it's only about uncontrolled components, but it touches on controlled components further down
- [`<GroupLabelInput />`](https://github.com/stride-nyc/remote_retro/blob/master/web/static/js/components/group_label_input.jsx)
  - at the moment, this component contains an uncontrolled input. That may need to change. Discuss what you think makes the most sense with your pair!
- [SemanticUI Icon Docs](https://semantic-ui.com/elements/icon.html)

PR #559, opened 2020-02-05
556

ajoseph-stride:
Added a Semantic UI checkbox upon successfully submitting a changed Group Label.  This checkbox appears next to the input field of the changed Group Label and fades out after two seconds.

Added additional Enzyme unit test to validate the appearance of the checkbox after successfully submitting the Group Label change.

__Relevant github Issue:__
- [issue (556)](https://github.com/stride-nyc/remote_retro/issues/556)


vanderhoop:
@ajoseph-stride! I see you removed the `mix.lock` file in 696ec0d0b43d8143bf91a287f14c69bcaf793eb5. The `*.lock` files are conventionally kept in version control, as these ensure that all developers on a team have the exact same node and elixir packages installed on their machines, respectively.

I suspect you removed the `mix.lock` file when trying to debug the chromedriver/chrome version descrepancy, but we'll want to reset the last commit to keep that file version control.

web/static/js/components/css_modules/group_label_input.css
```diff
@@ -33,6 +33,24 @@
   }
 }
 
+.update-succeeded-checkmark {
+  animation: fadeOut ease 2s;
+  -webkit-animation: fadeOut ease 2s;
+  -moz-animation: fadeOut ease 2s;
+  -o-animation: fadeOut ease 2s;
+  -ms-animation: fadeOut ease 2s;
```
vanderhoop: Thank you for adding these vendor prefixes, @JennMiller and @ajoseph-stride, but we can kill them, as we use a webpack plugin called `autoprefixer` that augments the shorthand declaration—in this case, `animation`—with all revelant vendor prefixes.

web/static/js/components/group_label_input.jsx
```diff
@@ -7,23 +7,63 @@ import styles from "./css_modules/group_label_input.css"
...
-        actions.submitGroupLabelChanges(groupWithAssociatedIdeasAndVotes, e.target.value)
-      }}
-    />
-    <div className="instruction">
-      <span className="keyboard-key">tab</span> to submit
-    </div>
-  </div>
-)
+class GroupLabelInput extends Component {
+  state = (() => {
```
vanderhoop: ahhh, given the destructuring you need to do here, it makes sense to house this in a es6 `constructor` function rather than in an immediately invoked function, as a) that'll pass in props as an argument, and b) it's more idiomatic of react class code when logic is necessary upon instantiation/construction. something like:

```es6
constructor(props) {
  const { groupWithAssociatedIdeasAndVotes: { label } } = props

  this.state = {
    oldLabel: label,
    showCheckmarkForLabelId: null,
  }
}
```

web/static/js/components/group_label_input.jsx
```diff
@@ -7,23 +7,63 @@ import styles from "./css_modules/group_label_input.css"
...
+  })();
+
+  static getDerivedStateFromProps(props, state) {
+    const { oldLabel } = state
+    const { groupWithAssociatedIdeasAndVotes } = props
+    const newLabel = groupWithAssociatedIdeasAndVotes.label
+    if (oldLabel !== newLabel) {
+      return ({
+        oldLabel: newLabel,
+        showCheckmarkForLabelId: groupWithAssociatedIdeasAndVotes.id })
```
vanderhoop: I think we can get away with changing `showCheckmarkForLabelId` -> `showCheckmarkForLabel` (Boolean), and assigning `true` in this case, and setting `false` in the constructor, as the `group`'s id will be unique to this instance.

web/static/js/components/group_label_input.jsx
```diff
@@ -7,23 +7,63 @@ import styles from "./css_modules/group_label_input.css"
...
+          defaultValue={groupWithAssociatedIdeasAndVotes.label}
+          maxLength={MAX_LENGTH_OF_GROUP_NAME}
+          onBlur={e => {
+            actions.submitGroupLabelChanges(groupWithAssociatedIdeasAndVotes, e.target.value)
+          }}
+        />
+        <div className="instruction">
+          <span className="keyboard-key">tab</span> to submit
+        </div>
+        {showCheckmarkForLabelId === groupWithAssociatedIdeasAndVotes.id
```
vanderhoop: per my note above, the group id will be unique to the instance of this component, so i think we can simplify the state attribute to a boolean at `showCheckmarkForLabel`, and simplify the render code to a short-circuit, like so: {`showCheckmarkForLabel && this.labelUpdatedCheckmark()`}

web/static/js/components/group_label_input.jsx
```diff
@@ -7,23 +7,63 @@ import styles from "./css_modules/group_label_input.css"
...
+      return ({
+        oldLabel: newLabel,
+        showCheckmarkForLabelId: groupWithAssociatedIdeasAndVotes.id })
+    }
+    return null
+  }
+
+  labelUpdatedCheckmark = () => {
+    const classes = ["check icon"]
+    classes.push(styles.updateSucceededCheckmark)
```
vanderhoop: given that `styles.updateSucceededCheckmark` maps to a string, we can remove the `Array.prototype.push` call in favor of an array literal like 

```es6
const classes = ["check icon", styles.updateSucceededCheckmark]
```

web/static/js/components/group_label_input.jsx
```diff
@@ -7,23 +7,63 @@ import styles from "./css_modules/group_label_input.css"
...
+    setTimeout(() => {
```
vanderhoop: i think there's a bug here, in that this `labelUpdatedCheckmark ` function could fire multiple times due to the react render cycle firing for reasons outside this particular component's control, i.e. a parent component's re-renders could cause this component to re-render multiple times a second, causing this function to be re-invoked multiple times, causing potential race conditions with the number of `setTimeouts` that get run

web/static/js/components/group_label_input.jsx
ajoseph-stride: Cool, didn't know that.  Modifying as per suggestion.

web/static/js/components/css_modules/group_label_input.css
ajoseph-stride: Thanks, didn't know that.  Modifying as per suggestion.

web/static/js/components/group_label_input.jsx
ajoseph-stride: We initially added the label id check since we found that all components were being updated when only one's label was modified.  However, with the new code, this no longer happens.  Thanks.  Modifying as per suggestion.

web/static/js/components/group_label_input.jsx
ajoseph-stride: We initially added the label id check since we found that all components were being updated when only one's label was modified.  However, with the new code, this no longer happens.  Thanks.  Modifying as per suggestion.

web/static/js/components/group_label_input.jsx
ajoseph-stride: Cool, didn't know any of that.  Modifying as per suggestion.

web/static/js/components/group_label_input.jsx
ajoseph-stride: Thanks for letting me know of this issue.  I am not sure how to properly fix the problem.  But, I'd love to pair with you as per your other comment to resolve this.

web/static/js/components/group_label_input.jsx
```diff
@@ -7,23 +7,60 @@ import styles from "./css_modules/group_label_input.css"
...
-)
+class GroupLabelInput extends Component {
+  constructor(props) {
+    super(props)
+    const { groupWithAssociatedIdeasAndVotes: { label } } = props
+    this.state = {
+      oldLabel: label,
+      showCheckmark: false,
+    }
+  }
```
JennMiller: ah, great! this makes sense!

web/static/js/components/group_label_input.jsx
ajoseph-stride: We addressed this as an unlikely scenario, but may merit investigation in the future.

Issue #595, opened 2021-11-09
[BUG] 500 after redirection from `Sign in with Google` button


manumc:
I was able to select my google account and sign in but on redirection back to `https://remoteretro.org/` I got a 500.

It _fix_ itself by just reloading the page on the browser.

I was not able to reproduce it given that once you login you can only logout but not delete your own account. Seems like is something that only happens first time.

Hope this is helpful :heart:

manumc:
PS. labels doesn't show up so I put it on the title

omgitsbillryan:
This is happening again 😄 
![image](https://github.com/stride-nyc/remote_retro/assets/3077884/91f12387-cabd-4c5e-a494-54adbd58b650)

omgitsbillryan:
I'm not able to fix it with a refresh, though 🤔

mwytock0812:
Hi @omgitsbillryan, Google stopped sending us some required params on auth success, which was causing 500s. This should be resolved. Please reopen if you're still seeing this.

PR #623, opened 2024-06-13
Make locale an optional user field to fix auth error

mwytock0812:
- Fixes issue with new users authenticating with Google OAuth getting a 500. Google stopped sending locale, which was a required field. Now we supply a default as a hotfix.

- Local testing replicates the 500 similar to #595 and the default value resolves it.

- #595


melissawahnish (approved):
:shipit: 
Thanks so much for doing this!

PR #627, opened 2024-06-14
Bust cache and update pins in phoenix buildpack

mwytock0812:
- Busts cache in Phoenix buildpack ([docs](https://github.com/gigalixir/gigalixir-buildpack-phoenix-static?tab=readme-ov-file#configuration))
- Pins the npm and yarn versions to those that work locally for production build
- Synchronizes local dev node version with buildpack version

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 erlang 22.3.4.16
 elixir 1.11.4
-nodejs 20.2.0
+nodejs 20.14.0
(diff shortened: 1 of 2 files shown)


vanderhoop (approved):
The version alignments here are 💯 . clean_cache=true might slow down deploys long-term, but that's a worry for after the locale-fix deploy goes through. Merge it!

PR #628, opened 2024-06-14
Revert all non-bug-fix changes made to application since last healthy deploy; use forked buildpack

vanderhoop:

This pull request is a squashed reversion of all application code changes made to the app since the last healthy deploy (in January of 2023!), save for the locale bugfix of #623, which I've kept. The one substantive change I've made is to the node buildpack, which I've updated to a fork I use for another production app on Gigalixir.

This PR should be merged with a squash commit, as that will provide an ease-of-reversion, as a) it might not work when deploying to your prod env, and b) you'll probably want to revisit those dependency upgrades at a later date anyway.


mwytock0812 (approved):
Approving and merging this.

PR #632, opened 2025-02-11
Update node to 22+

jrdrew:

Updated the node version 22

### DIFF
--- a/bin/install_node_with_dependencies
+++ b/bin/install_node_with_dependencies
@@ -1,10 +1,10 @@
-#1/bin/bash
+#!/bin/bash
 
 # install nvm
 brew install nvm
 
-# install project version of node (18.19.0) and use it
-nvm install 18.19.0 && nvm use 18.19.0
+# install project version of node (22.0.0) and use it
+nvm install 22.0.0 && nvm use 22.0.0
 
 # install yarn to install package
 npm install yarn -g
--- a/bin/reset_dev
+++ b/bin/reset_dev
@@ -1,8 +1,4 @@
 #!/bin/bash
 
 # clean existing build/dependencies and get/compile
-rm -rf _build/
-mix deps.clean --all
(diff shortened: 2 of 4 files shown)


bin/install_node_with_dependencies
```diff
@@ -1,13 +1,13 @@
-#1/bin/bash
+#!/bin/bash
 
 # install nvm
 brew install nvm
 
-# install project version of node (18.19.0) and use it
-nvm install 18.19.0 && nvm use 18.19.0
+# install project version of node (22.0.0) and use it
+nvm install 22.0.0 && nvm use 22.0.0
```
dvndrsn, line 7: Does `nvm` use `.tool-versions` file? I believe `asdf` introduced this file and is part of the instructions for managing erlang and elixir.

MattM-StrideNYC: NVM would use an `.nvmrc` file. We no longer need the `.tool-versions` file since we eliminated ASDF.

bin/reset_dev
```diff
@@ -1,8 +1,4 @@
 #!/bin/bash
 
 # clean existing build/dependencies and get/compile
-rm -rf _build/
-mix deps.clean --all
-mix deps.get
-mix deps.compile
-mix compile
\ No newline at end of file
+rm -rf _build/ && mix deps.clean --all && mix deps.get && mix deps.compile && mix compile
```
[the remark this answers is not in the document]
MattM-StrideNYC: Good question. Probably an auto-formatter/linter making this change, I would guess?

The only difference I see is that the new version is dependent on the former steps/invocations completing successfully before executing the next one in sequence. The original would still execute the statements regardless of whether or not the prior one(s) completed successfully or not. Generally, that's not what we want I would think.

StrideJoshKK:
Stamped!

PR #657, opened 2025-04-03
Rewrite React DND (Drag'N Drop) to use React hooks - Idea Column Stage

paytown:
Updates the idea column stage to use `dnd-kit`. It also untangles and removes some components that IMO were unnecessarily complex in order to make them easier to maintain.

Unit tests written in Jest, all passing

Screenshot 2025-04-08 at 12 45 54 PM

Screenshot 2025-04-08 at 12 47 45 PM


config/dev.exs
```diff
@@ -46,7 +46,7 @@ config :remote_retro, RemoteRetro.Repo,
   username: "postgres",
   password: "postgres",
   database: "remote_retro_dev",
-  hostname: "db",
+  hostname: "localhost",
```
paytown, line 49: TODO: Roll back

web/static/js/components/draggable_idea_content.jsx
```diff
@@ -1,12 +1,16 @@
+// TODO: Test if we can delete this file
```
paytown, line 1: Yes we can delete this file. We need to remove the IdeaGroupingCard before doing this however. That is done in the grouping PR.

MattM-StrideNYC (approved):

PR #659, opened 2025-04-10
Grouping board - in progress drag updates for all users live

paytown:
Updates the GroupingBoard and GroupingCard to move for non dragging users when a user is dragging a card. Also implements performance improvements such as memoization and batching measurement calculations.

- Manual testing in browser
- AI assisted unit tests


test/components/grouping_board.test.js
```diff
@@ -48,7 +89,6 @@ describe("GroupingBoard", () => {
 
         const ideaCards = screen.getAllByTestId("grouping-card")
         expect(ideaCards).toHaveLength(2)
-        // Check order by data-idea-id attribute
```
MattM-StrideNYC, line 51: Is there a reason we need to remove a lot of these comments? They seem to be still applicable, is that not the case?

web/static/js/components/grouping_card.jsx
```diff
@@ -75,11 +87,23 @@ const GroupingCard = forwardRef(
...
   idea: AppPropTypes.idea.isRequired,
-  isActive: PropTypes.bool,
   currentUser: AppPropTypes.presence.isRequired,
   groupId: PropTypes.number,
   userOptions: AppPropTypes.userOptions.isRequired,
   children: PropTypes.node.isRequired,
 }
 
-export default GroupingCard
+const MemoizedGroupingCard = React.memo(GroupingCard, (prevProps, nextProps) => {
```
MattM-StrideNYC, line 96: Is this used elsewhere? I don't see it used anywhere in the PR

web/static/js/services/idea_card_grouping.js
```diff
@@ -47,6 +43,18 @@ const areElementsOverlapping = (rect1, rect2) => {
   )
 }
 
+const batchGetBoundingClientRects = cardRefs => {
```
MattM-StrideNYC, line 46: I am curious, what is the benefit of this method over doing the original implementation?:

```js
    const rects = {}
    cardIds.forEach(id => {
      const ref = cardRefs[id]
      rects[id] = ref?.getBoundingClientRect()
    })
```

test/components/grouping_board.test.js
paytown: Oh they're still applicable, it was just general cleanup because it seems like elsewhere in the codebase we prefer less vs. more comments. I can return them if you prefer.

web/static/js/components/grouping_card.jsx
paytown: Exporting it as a default will let you import it as GroupingCard by the GroupingBoard component. I see the confusion though and will update.

test/components/grouping_board.test.js
MattM-StrideNYC: Let's add them back. Comments are useful and we should keep them around (and update them if needed).

web/static/js/components/grouping_card.jsx
MattM-StrideNYC: Ah, I probably misread it.

web/static/js/services/idea_card_grouping.js
paytown: Honestly there's not much improvement, I think I was just getting too clever trying to refactor. I've returned the original code with some memoization in place to hopefully help optimize perf.. Lmk what you think.

test/components/grouping_board.test.js
paytown: Returned where it makes sense. Going to mark as resolved.

web/static/js/components/grouping_card.jsx
paytown: All good, it wasn't really evident what was going on. I prefer the update anyway, makes it more legible. Going to mark as resolved.

MattM-StrideNYC (approved):

